Issue #25, opened 2021-11-10
Lint and Test Workflow


cmarshak:
- copy simple test/lint github action based on:
    - this for [linting](https://github.com/ASFHyP3/hyp3-sdk/blob/develop/.github/workflows/static-analysis.yml#L1-L23)
    - this for [testing](https://github.com/ASFHyP3/hyp3-sdk/blob/develop/.github/workflows/test-and-tag.yml#L1-L34)

cmarshak:
#26 fixes

PR #26, opened 2021-11-10
Test and lint workflow

cmarshak:
Addresses Issue #25.

### DIFF
--- a/isce2_topsapp/__init__.py
+++ b/isce2_topsapp/__init__.py
@@ -5,3 +5,12 @@ from .localize_aux_cal import download_aux_cal
 from .topsapp_proc import topsapp_processing
 from .packaging import package_gunw_product
 from .delivery_prep import prepare_for_delivery
+
+
+__all__ = ['download_orbits',
+           'download_slcs',
+           'download_dem_for_isce2',
+           'download_aux_cal',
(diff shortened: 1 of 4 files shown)


isce2_topsapp/__init__.py
```diff
@@ -5,3 +5,12 @@
...
 from .delivery_prep import prepare_for_delivery
+
+
+__all__ = ['download_orbits',
+           'download_slcs',
+           'download_dem_for_isce2',
+           'download_aux_cal',
+           'topsapp_processing',
+           'package_gunw_product',
+           'prepare_for_delivery']
```
cmarshak, line 16: I added this for flake8 and is based on what hyp3 repos do.

jhkennedy (approved):
This all looks fine to me. 

You're running pytest for *every* push, to any branch, and for any PR that's opened, which is fine but possibly excessive. We tend to split static analysis to it's own workflow so it always runs, and pytest to run just on pushes/PR so `dev` and `main` as other pushes/PR are likely to be part of active development and may be broken

cmarshak:
I am going to leave this broad workflows being trigger for now and put an issue so this doesn't get left out. Thank you, Joe, for reviewing!

PR #29, opened 2021-11-15
Update JSON message for DAAC Ingest

### LINKED ISSUE #16: Update json file to match with delivery schema
Fix the JSON output file to match the schema here: https://github.com/asfadmin/grfn-ingest/blob/test/verify/src/message_schema.json

Create a test to ensure correct output in test case.

First created here: https://github.com/ACCESS-cloud-based-InSAR/DockerizedTopsApp/pull/10

cmarshak:
Fixes #16


cmarshak:
Until we migrate fully to pytest or github integrates nbdiff more meaningfully, just updated a few cells in the test notebook for delivery inspection most importantly

```
schema = get_dataset_schema()
metadata = json.load(open('S1-GUNW-A-R-064-tops-20210723_20210711-015001-35393N_33512N-PP-6267-v2_0_4/dataset.json'))

assert(jsonschema.validate(instance=metadata, schema=schema) is None)
```

jhkennedy:
> github integrates nbdiff more meaningfully

@cmarshak you might want to check out https://www.reviewnb.com/ -- free for open source

tests/prepare-for-delivery.ipynb
```diff
@@ -3,11 +3,11 @@
   {
```
cmarshak, line 3: <p>Interesting... I wonder how this appears on github.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/pull/29/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

isce2_topsapp/delivery_prep.py
```diff
@@ -115,26 +124,30 @@ def format_metadata(nc_path: Path,
...
+                     "sensing_start": f'{ref_props["startTime"]}Z',
+                     "sensing_stop": f'{ref_props["stopTime"]}Z',
+                     "orbit_number": [int(ref_props['orbit']),
+                                      int(sec_props['orbit'])],
+                     "platform": [ref_props['platform'], sec_props['platform']],
                      "beam_mode": ref_props['beamModeType'],
                      "orbit_direction": ref_props['flightDirection'].lower(),
                      "dataset_type": 'slc',
                      "product_type": 'interferogram',
                      "polarization": "HH",
```
asjohnston-asf, line 142: Possibly out of scope for this PR, but I'd expect polarization to be determined from the input SLC(s) rather than being hardcoded.

isce2_topsapp/delivery_prep.py
```diff
@@ -115,26 +124,30 @@ def format_metadata(nc_path: Path,
...
+    ogr_bbox = all_metadata['intersection_geo'].envelope.exterior.coords[:4]
+    metadata.update({"ogr_bbox": ogr_bbox,
                      "reference_scenes": all_metadata['reference_scenes'],
                      "secondary_scenes": all_metadata['secondary_scenes'],
-                     "sensing_start": ref_props['startTime'],
-                     "sensing_stop": ref_props['stopTime'],
-                     "orbit_number": [ref_props['orbit'], sec_props['orbit']],
-                     "platform": ref_props['platform'],
+                     "sensing_start": f'{ref_props["startTime"]}Z',
+                     "sensing_stop": f'{ref_props["stopTime"]}Z',
```
asjohnston-asf, line 134: These values will be set as the start/end times for the GUNW product in CMR and Vertex. I suspect searching for GUNW products would be more intuitive if these were the start of the earliest scene and the stop of the latest scene, rather than only the start/stop of the reference scene. I don't search for interferograms often myself, though, so it's likely worth gathering more input before making changes.

asjohnston-asf (dismissed):

asjohnston-asf:
Metadata changes look good. I built the container from this branch and re-generated [S1-GUNW-D-R-006-tops-20211204_20211122-030141-09305N_07267N-PP-2bce-v2_0_4](https://search.asf.alaska.edu/#/?searchType=List%20Search&searchList=S1-GUNW-D-R-006-tops-20211204_20211122-030141-09305N_07267N-PP-2bce-v2_0_4&resultsLoaded=true&granule=S1-GUNW-D-R-006-tops-20211204_20211122-030141-09305N_07267N-PP-2bce-v2_0_4-amplitude&zoom=6.145&center=41.477,2.996) using the updated code. Metadata file passes validation and values match up with the current product in Vertex.

I left comments regarding the polarization and start/stop time attributes, but I wouldn't consider those blocking at this time.

isce2_topsapp/delivery_prep.py
cmarshak: I totally agree and thought about this myself. Unfortunately, the existing products do not do this, which means searching using these metadata fields would be inconsistent and even more annoying.

https://github.com/aria-jpl/topsApp_pge/blob/develop/create_standard_product_s1.py#L459-L478

If you have a suggestion on how to remedy this previous oversight or can propose a workflow to deal with this possible inconsistency - I am all ears.

dbekaert:
Agreed, not a backward compatible change.
If it is trivial to touch the past products in CMR, then might be nice to
get it corrected?
Given the SAR repeat pass behaviour the times (HH:MM) itself are not much
different.
The end time could be updated by replacing the end date YYYMMDD itself,
keeping the time the same.
One would need to check IFGs on the date line transition (believe we have
some for Tibet) for any date changes.

I recall this feature does not currently work in the python API search
either.
We actually have a manual fix in the ARIA-tools download that wraps around
the API and does some extra filtering.

asjohnston-asf:
Updating the dates for existing products in CMR might be hours to days of work, so neither trivial or terrible. I'll defer to @dbekaert and @jhkennedy on whether that's worth pursuing and when.

isce2_topsapp/delivery_prep.py
cmarshak: Wow - that's a very good observation.

I actually don't know precisely - I believe it's a topsapp default. I just pass the entire zip file as here: https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/dev/isce2_topsapp/templates/topsapp_template.xml

And then the ISCE2 topsapp I believe just uses HH.

Issue #19, opened 2021-11-09
Browse Imagery


cmarshak:
- Erode/dilate water mask for cleaner browse image: https://github.com/ACCESS-cloud-based-InSAR/DockerizedTopsApp/pull/10

cmarshak:
Currently, I think it's best to use ASF's water mask. It will be much faster and cleaner IMHO.

I would do something like add a `buffer` parameter to this [function](https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/9714bb712d49e680a13cfabbb8891fd0afad136e/isce2_topsapp/water_mask.py#L90-L119) using shapely's `buffer`. You could use `ndimage`, but this will be less elegant since with the vectors you don't have to worry about a mask.

Then make sure it is called with correct buffer [here](https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/dev/isce2_topsapp/delivery_prep.py#L86).

There is an example notebook currently: https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/dev/tests/prepare-for-delivery.ipynb (I am going to move these into separate notebooks for future references and preserve tests). I may move this notebook into something like "notebooks/". But hopefully, this helps you get started.

There is basically removing the 3km buffer that is referenced in #10.

If this doesn't work, we can figure out an alternative with ISCE2.

cmarshak:
@jhkennedy - just to confirm - you do not have the original USGS water mask available for download, right? The [data](https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/9714bb712d49e680a13cfabbb8891fd0afad136e/isce2_topsapp/water_mask.py#L56-L57) you derived this water mask from is on a server but zipped (so we have to totally unzip it to read a bbox).

PR #34, opened 2021-12-02
Update water mask with full resolution GMT mask

sssangha:
Now use full res, land + ocean boundary mask uploaded to S3 by @jhkennedy.

Addresses #19

### DIFF
--- a/isce2_topsapp/water_mask.py
+++ b/isce2_topsapp/water_mask.py
@@ -53,8 +53,8 @@ def rasterize_shapes_to_array(shapes: list,
 
 
 def get_water_mask_dataframe(bounds: list) -> gpd.GeoDataFrame:
-    mask_location = ('/vsicurl/https://asf-dem-west.s3.amazonaws.com/'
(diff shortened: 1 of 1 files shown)


cmarshak:
Can you share what the browse imagery will look like with this update particularly near a water body. Ideally the Hawaii example that @gracebato shared - https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/pull/10#issuecomment-940374130.

Or somewhere else. Just to demonstrate to David that there is no coastline artifacts 🤞 

Thanks, Sim!

sssangha:
@cmarshak yes no problem. It's taking a while to generate the figures for Hawaii, so in the meantime here they are for SoCal:
CA_mask_zoom
CA_mask

sssangha:
Actually, Hawaii just finished up
Hawaii_mask
!

cmarshak (approved):
Beautiful!

dbekaert:
Looks great.

Thnx

PR #42, opened 2022-01-24
Fix collections.Iterable error

jhkennedy:
Products being created successfully but this error
image

is being ignored when packaging due to this exception handeling:
https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/02d2934794767038c2fe905616acf18d86bffaf8/isce2_topsapp/packaging_utils/nc_packaging.py#L601-L608


cmarshak (commented):
Awesome catch.

Packaging will have errors but will continue - thanks for catching this.

PR #50, opened 2022-02-04
Allow Earthdata credentials via environment variables

jhkennedy:
HyP3 is moving to provide Earthdata credentials to the container via environment variables instead of CLI options to keep them encrypted throughout the processing chain.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -5,23 +5,24 @@ channels:
  - defaults
 dependencies:
  - python>=3.8
- - isce2
- - jupyter
- - papermill
- - ipykernel
- - rasterio
- - shapely
+ - pip
+ - asf_search>=3.0.4
+ - boto3
  - dateparser
- - jinja2
- - numpy
- - matplotlib
- - netcdf4
  - fiona
  - geopandas
- - pip
- - boto3
- - asf_search>=3.0.4
+ - ipykernel
+ - isce2
+ - jinja2
  - jsonschema==3.2.0
+ - jupyter
+ - matplotlib
+ - netcdf4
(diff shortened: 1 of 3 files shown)


environment.yml
```diff
@@ -5,23 +5,24 @@ channels:
...
- - jinja2
- - numpy
- - matplotlib
- - netcdf4
- - fiona
- - geopandas
- - pip
- - boto3
- - asf_search>=3.0.4
- - jsonschema==3.2.0
```
jhkennedy, line 24: alphabatized and added pytest

isce2_topsapp/__main__.py
```diff
@@ -37,33 +39,54 @@ def localize_data(reference_scenes: list,
...
+    except (netrc.NetrcParseError, TypeError):
+        raise ValueError(
+            f'Please provide valid Earthdata login credentials via {netrc_file}, '
+            f'the --username and --password CLI option, or '
+            f'the EARTHDATA_USERNAME and EARTHDATA_PASSWORD environment variables.'
+        )
+
+    return username, password
+
+
```
cmarshak, line 73: Beautiful! So, all the readme instructions about netrc are valid and so are the command line scripts that use --username/--password. Thanks Joe.

cmarshak:
I will approve after it passes the CI/CD.

asjohnston-asf (approved):

isce2_topsapp/__main__.py
jhkennedy: yeah! Should just extend functionality

PR #51, opened 2022-03-11
Use hyp3lib for orbits

jhkennedy:
A lot of scenes in the Oklahoma track failed due to not having all the necessary orbit information

This 
* uses `hyp3lib` to download orbits which adds, and prefers, the ESA orbit API and falls back on the ASF orbit heap. 
* reworks the orbits to fetch the appropriate orbits for every scene (instead of just the first reference and secondary)
* points ISCE to the orbit directory instead of specific orbit files


jhkennedy:
I was was able to run the standard example from the README as well as one of the Texas scenes that failed due to orbits. Should be g2g.

dbekaert (approved):
LGTM

PR #54, opened 2022-04-01
install released dem-sticher, not dev version

jhkennedy:
This will ensure more reproducible builds, at the expense of needed to release dem-sticher to incorporate changes


jhkennedy:
Closing as this is already done, but better:
https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/dev/environment.yml#L40

PR #73, opened 2022-10-07
Enable Processing of Single Burst Interferograms

forrestfwilliams:
# Purpose
As ASF's Ingest & Access team works to create a service that can provide users a way to download Sentinel-1 SLC data for a single burst, we want to create ways for users to create single-burst interferograms. This PR adds functionality and a new entrypoint to `DockerizedTopsApp` that allows it to interact with the alpha-stage burst API and create single-burst interfograms.

#  Additions
What this PR adds:

- Utilities to interact with the burst API
- Utilities to download metadata and swath data not currently supported by the API
- An alternate entrypoint that allows for burst processing
- Tests for these utilities

What this PR does NOT add:
- A stable way to interact with the burst API
- Packaging for the burst products beyond what is output from ISCE2

# Comparison Data
This [link](https://ffwilliams2-shenanigans.s3.us-west-2.amazonaws.com/bursts/burst_comparison.zip) contains data that can be used to compare the outputs generated by ISCE2 for an entire SLC, and the new single-burst option within `DockerizedTopsApp`. Data in the `bursts` directory ISCE2 geocoded single-burst outputs from the 2020 Iranian and the 2020 Greecian earthquake. The `slc` directory contains full slc outputs for the same test cases.

Within a correctly configured `DockerizedTopsApp` environment, the burst test cases can be run using these commands:

For the Iranian test case:
```bash
gunw_burst --username $EDL_USERNAME \
           --password $EDL_PASSWORD \
           --reference-scene S1A_IW_SLC__1SDV_20200604T022251_20200604T022318_032861_03CE65_7C85 \
           --secondary-scene S1A_IW_SLC__1SDV_20200616T022252_20200616T022319_033036_03D3A3_5D11 \
           --image-number 5 \
           --burst-number 8 \
           --azimuth-looks 4 \
           --range-looks 20
```

For the Greecian test case:
```bash
gunw_burst --username $EDL_USERNAME \
           --password $EDL_PASSWORD \
           --reference-scene S1B_IW_SLC__1SDV_20201115T162313_20201115T162340_024278_02E29D_5C54 \
           --secondary-scene S1A_IW_SLC__1SDV_20201203T162353_20201203T162420_035524_042744_6D5C \
           --image-number 5 \
           --burst-number 1 \
           --azimuth-looks 4 \
           --range-looks 20
```

The SLC test cases were run directly with `topsApp.py` using the `topsApp.xml` configuration files that were generated for each respective burst test case where the `swath` and `region of interest` paramaters had been commented out. These configuration files can be found in the same directory as this readme.

--- 

# On the entry_points

This makes what should be completely transparent changes to the DockerizedTopsApp entrypoint.

Previously, either via `docker run` or `python -m isce2_topsapp` the `main` entrypoint would look like:
```bash
$ python -m isce2_topsapp --help
usage: gunw_slc [-h] [--username USERNAME] [--password PASSWORD] [--bucket BUCKET] [--bucket-prefix BUCKET_PREFIX] [--dry-run] --reference-scenes
                REFERENCE_SCENES [REFERENCE_SCENES ...] --secondary-scenes SECONDARY_SCENES [SECONDARY_SCENES ...]

optional arguments:
  -h, --help            show this help message and exit
  --username USERNAME
  --password PASSWORD
  --bucket BUCKET
  --bucket-prefix BUCKET_PREFIX
  --dry-run
  --reference-scenes REFERENCE_SCENES [REFERENCE_SCENES ...]
  --secondary-scenes SECONDARY_SCENES [SECONDARY_SCENES ...]
```
In this PR, that behavior stays intact, but we use a argparser and entrypoints trick to allow alternate entrypoints. Instead, if you run:
```bash
$ python -m isce2_topsapp ++help
usage: __main__.py [+h] [++process {gunw_slc,gunw_burst}]

optional arguments:
  +h, ++help            show this help message and exit
  ++process {gunw_slc,gunw_burst}
                        Select the HyP3 entrypoint to use (default: gunw_slc)
```
You can see how to select an alternative entrypoint. So this will run the GUNW process on full SLCs (the default, and only previously):
```bash
$ # same as python -m isce2_topsapp --help
$ python -m isce2_topsapp ++process gunw_slc --help
usage: gunw_slc [-h] [--username USERNAME] [--password PASSWORD] [--bucket BUCKET] [--bucket-prefix BUCKET_PREFIX] [--dry-run] --reference-scenes
                REFERENCE_SCENES [REFERENCE_SCENES ...] --secondary-scenes SECONDARY_SCENES [SECONDARY_SCENES ...]

optional arguments:
  -h, --help            show this help message and exit
  --username USERNAME
  --password PASSWORD
  --bucket BUCKET
  --bucket-prefix BUCKET_PREFIX
  --dry-run
  --reference-scenes REFERENCE_SCENES [REFERENCE_SCENES ...]
  --secondary-scenes SECONDARY_SCENES [SECONDARY_SCENES ...]
  ```
  and this will instead invoke the single burst processing entrypoint:
  ```bash
  $ python -m isce2_topsapp ++process gunw_burst --help
  usage: gunw_burst [-h] [--username USERNAME] [--password PASSWORD] [--bucket BUCKET] [--bucket-prefix BUCKET_PREFIX] [--dry-run] --reference-scene
                  REFERENCE_SCENE --secondary-scene SECONDARY_SCENE --image-number IMAGE_NUMBER --burst-number BURST_NUMBER
                  [--azimuth-looks AZIMUTH_LOOKS] [--range-looks RANGE_LOOKS]

optional arguments:
  -h, --help            show this help message and exit
  --username USERNAME
  --password PASSWORD
  --bucket BUCKET
  --bucket-prefix BUCKET_PREFIX
  --dry-run
  --reference-scene REFERENCE_SCENE
  --secondary-scene SECONDARY_SCENE
  --image-number IMAGE_NUMBER
  --burst-number BURST_NUMBER
  --azimuth-looks AZIMUTH_LOOKS
  --range-looks RANGE_LOOKS
  ```
  
  tl;dr: By adding a `++process` argument you can select a different entrypoint.

### CODE REVIEW

**Comment by jhkennedy** [comment]
Note tests are failing due to being a fork because:
* Pytest is failing because it doesn't have access to the ACCESS-Cloud-Based-InSAR secrets
* Docker build is failing because it can't push to ACCESS-Cloud-Based-InSAR container repository

File: isce2_topsapp/__main__.py [hunk 1]
```diff
@@ -129,5 +133,84 @@ def main():
...
+    parser.add_argument('--secondary-scene', type=str, required=True)
+    parser.add_argument('--image-number', type=int, required=True)
+    parser.add_argument('--burst-number', type=int, required=True)
+    parser.add_argument('--azimuth-looks', type=int, default=2)
+    parser.add_argument('--range-looks', type=int, default=10)
+    args = parser.parse_args()
+
+    ensure_earthdata_credentials(args.username, args.password)
+
+    ref_resp, sec_resp = get_asf_slc_objects([args.reference_scene, args.secondary_scene])
```
jhkennedy, line 153:
```suggestion
    ref_obj, sec_obj = get_asf_slc_objects([args.reference_scene, args.secondary_scene])
```
I think `resp` here is response? Generally, I'd like the variables to match the fn signature and it's typically worth spelling things all the way out.

isce2_topsapp/__main__.py
```diff
@@ -129,5 +133,84 @@ def main():
...
+    sec_params = BurstParams(
+        safe_url=sec_resp.properties['url'],
+        image_number=args.image_number,
+        burst_number=args.burst_number,
+    )
+
+    ref_burst, sec_burst = download_bursts([ref_params, sec_params])
+
+    intersection = ref_burst.footprint.intersection(sec_burst.footprint).bounds
+    asc = ref_burst.orbit_direction == 'ascending'
```
jhkennedy, line 169:
```suggestion
    is_ascending = ref_burst.orbit_direction == 'ascending'
```
would make the intent of this variable much clearer.

isce2_topsapp/__main__.py
```diff
@@ -129,5 +133,84 @@ def main():
...
+    roi = get_region_of_interest(ref_burst.footprint, sec_burst.footprint, asc)
+
+    out_orbits = download_orbits([ref_burst.safe_name[:-5]], [sec_burst.safe_name[:-5]], dry_run=args.dry_run)
```
jhkennedy, line 172:
```suggestion
    orbits = download_orbits([ref_burst.safe_name[:-5]], [sec_burst.safe_name[:-5]], dry_run=args.dry_run)
```
not sure the "out" adds much value to the name

isce2_topsapp/__main__.py
```diff
@@ -129,5 +133,84 @@ def main():
...
+
+    if not args.dry_run:
+        # TODO this is likely not the optimal geometry to pass to this function
+        out_dem = download_dem_for_isce2(intersection)
```
jhkennedy, line 176:
```suggestion
        dem = download_dem_for_isce2(intersection)
```
likewise here

isce2_topsapp/localize_burst.py
```diff
@@ -0,0 +1,315 @@
...
+
+import pandas as pd
+import requests
+from shapely import geometry
+
+
+@dataclass
+class BurstParams:
+    """Class that contains the parameters nessecary to request a burst from the API."""
+
```
jhkennedy, line 17:
```suggestion
```
Space here is unnecessary I think

isce2_topsapp/localize_burst.py
```diff
@@ -0,0 +1,315 @@
...
+class BurstMetadata:
+    def __init__(self, metadata: ET.Element, manifest: ET.Element, burst_params: BurstParams):
+        self.safe_url = burst_params.safe_url
+        self.image_number = burst_params.image_number
+        self.burst_number = burst_params.burst_number
+        self.safe_name = Path(self.safe_url).with_suffix('.SAFE').name
+
+        image_numbers = [int(x.attrib['source_filename'].split('-')[-1][2]) for x in metadata]
+        products = [x.tag for x in metadata]
+        combos = list(zip(image_numbers, products))
```
jhkennedy, line 32: Not super in love with `combos` as I have to keep looking back to understand what's inside it. Not sure what a better name would be though...

isce2_topsapp/localize_burst.py
```diff
@@ -0,0 +1,315 @@
...
+        footprint = geometry.Polygon(zip(x, y))
+        centroid = tuple([x[0] for x in footprint.centroid.xy])
+        return footprint, footprint.bounds, centroid
+
+
+def create_burst_request(burst_params: BurstParams, content: str) -> dict:
+    urls = {
+        'metadata': 'https://g6rmelgj3m.execute-api.us-west-2.amazonaws.com/metadata',
+        'geotiff': 'https://g6rmelgj3m.execute-api.us-west-2.amazonaws.com/geotiff',
+    }
```
jhkennedy, line 91: I think I'd hardcode this at the top of the file in all caps instead of inside this function to make it clear it's constant/magic string.

isce2_topsapp/localize_burst.py
```diff
@@ -0,0 +1,315 @@
...
+    burst_request = create_burst_request(burst_params, content='metadata')
+    burst_request['cookies'] = {'asf-urs':asf_session.cookies['asf-urs']}
+
+    i = 1
+    downloaded = False
+    while (not downloaded) and (i <= 3):
+        print(f'Download attempt #{i}')
+        response = asf_session.get(**burst_request)
+        downloaded = response.ok
+        i += 1
```
jhkennedy, line 134:
```suggestion
    for ii in range(1, 4):
        print(f'Download attempt #{ii}')
        response = asf_session.get(**burst_request)
        if (downloaded := response.ok):
            break
```
while loop with an iterator is just a for loop.

isce2_topsapp/localize_burst.py
```diff
@@ -0,0 +1,315 @@
...
+        f.write(response.content)
+
+    return str(out_file)
+
+
+def download_manifest(safe_url: str, out_file: Union[Path, str] = None) -> ET.Element:
+    import netrc
+
+    import aiohttp
+    import fsspec
```
jhkennedy, line 149: I think we talked about this, but why aren't these top-level imports? Just because we expect to drop this function?

isce2_topsapp/localize_burst.py
```diff
@@ -0,0 +1,315 @@
...
+        asc: bool = True) -> Tuple[float, float, float, float]:
+    bbox1 = geometry.box(*poly1.bounds)
+    bbox2 = geometry.box(*poly2.bounds)
+    intersection = bbox1.intersection(bbox2)
+    bounds = intersection.bounds
+
+    x, y = (0, 1) if asc else (2, 1)
+    roi = geometry.Point(bounds[x], bounds[y]).buffer(0.005)
+    minx, miny, maxx, maxy = roi.bounds
+    return minx, miny, maxx, maxy
```
jhkennedy, line 276: Any reason to not just return `roi.bounds` and document the order they are returned in?

isce2_topsapp/localize_burst.py
```diff
@@ -0,0 +1,315 @@
...
+        'client_id': 'BO_n7nTIlMljdvU6kRRB3g',
+        'redirect_uri': 'https://auth.asf.alaska.edu/login',
+    }
+    response = session.get('https://urs.earthdata.nasa.gov/oauth/authorize', params=payload)
+    response.raise_for_status()
+    return session
+
+
+# TODO base_path is not used, can we remove it?
+def download_bursts(param_list: Iterator[BurstParams], base_path: Path = Path.cwd()) -> List[BurstMetadata]:
```
jhkennedy, line 294:
```suggestion
def download_bursts(param_list: Iterator[BurstParams]) -> List[BurstMetadata]:
```

isce2_topsapp/templates/topsapp_template.xml
```diff
@@ -17,7 +17,7 @@
     <property name="range looks">{{ range_looks }}</property>
     <property name="azimuth looks">{{ azimuth_looks }}</property>
     <property name="filter strength">{{ filter_strength }}</property>
-    <!-- <property name="region of interest">{{ region_of_interest }}</property> -->
+    <property name="region of interest">{{ region_of_interest }}</property>
```
jhkennedy, line 20: may need to do some jinja logic here as it was commented out from GUNW SLC generation. I think you can just:

```
{% if region_of_interest }<property name="region of interest">{{ region_of_interest }}</property>{% endif %}
```
though you may need to adjust the whitespace control: https://jinja.palletsprojects.com/en/3.1.x/templates/#whitespace-control

jhkennedy: Hmm, looks like they do specify it for GUNW SLCs and it might just work. @cmarshak why's region of interest commented out when you do compute it and pass it to the template? Looks like it was added that way initially so I don't have any insight there.

jhkennedy (commented):
Overall looks really good! I left a bunch of nit-picky-level comments and would approve with or without those.

There is one question about how the addition of the `region_of_interest` in the topssApp XML template could affect the GUNW SLC process, that should be addressed first though.

isce2_topsapp/localize_burst.py
forrestfwilliams: No, not really.

isce2_topsapp/localize_burst.py
forrestfwilliams: changed to `image_numbers_and_products`

isce2_topsapp/localize_burst.py
forrestfwilliams: Agreed. Will move.

isce2_topsapp/__main__.py
forrestfwilliams: I've changed this, as well as the argument name in the function it is used in.

isce2_topsapp/templates/topsapp_template.xml
cmarshak: It's from the original in the ariamh repo and I had messed around with it to reduce processing time for tests - but was annoyed that if it's included then, it cannot be left blank. Jinja would solve it

cmarshak: Honestly, I think it's fine like this: https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/9fd15a2beda86b347dd4f4c236cfd38a87b3c2e6/isce2_topsapp/topsapp_proc.py#L42-L57

PR #74, opened 2022-10-20
Adding Ionosphere Phase Estimation

cmarshak:
Provides API functionality (through the associated CLI of the plugin) for:

1. ionosphere phase estimation (to be used for ionosphere correction)
2. ESD correction

Both of which are included in ISCE2 and requires modification of the `topsApp.xml` file (look at the [example xml](https://github.com/isce-framework/isce2/blob/main/examples/input_files/topsApp.xml) for all the features that can be included). Here is our [template](https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/ionosphere/isce2_topsapp/templates/topsapp_template.xml) for this PR.

This pull request also packages the ionosphere layer within the GUWN product. Specifically, takes `merged/topophase.ion.geo` and puts into the netcdf group `/science/grids/corrections/derived/ionosphere`.

The ESD outputs are not preserved currently. This is because I am not aware how.

The current test for the PR is the following:

```
isce2_topsapp --reference-scenes S1A_IW_SLC__1SDV_20221002T151520_20221002T151543_045265_056931_E517 \
              --secondary-scenes S1A_IW_SLC__1SDV_20220908T151520_20220908T151542_044915_055D68_78EC \
              --estimate-ionosphere-delay True \
              --do-esd True \
              --esd-coherence-threshold .5 \
              > topsapp_img.out 2> topsapp_img.err
```

The above test is included in the readme.

*Additional notes*: 

+ I updated the inclusion of the ionosphere delay layer to use `xarray` (and `rioxarray`). It is much cleaner than writing netcdf attributes and coordinates directly. However, because there is a placeholder  for`/science/grids/corrections/derived/ionosphere` typed as a string in the existing packaging template and workflow, I simply removed this from the packaging template json [here](https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/dev/isce2_topsapp/templates/nc_packaging_template.json#L253-L275). Reproducing the json excerpt below for posterity as the link is to the dev branch, which will presumably be updated if/when this PR is merged. There might be better ways of overwriting variables within `xarray` (specifically overwriting their type), but haven't looked into it. As a result, if ionosphere is _turned off_, the layer will be deleted (no placeholder).
```
                      "name" : "corrections",
                      "content" : [
                        {
                          "group" : [
                            {
                              "name" : "derived",
                              "content" : [
                                {
                                  "group" : [
                                    {
                                      "name" : "ionosphere",
                                      "content" : [
                                      {
                                          "dataset" : [
                                          {
                                            "name" : "ionosphere",
                                            "type" : "str",
                                            "description" : "placeholder dummy: [2D]"
                                          }]
                                      }]
                                    }]
                                }]
                            }
```
+ The application of the ionosphere correction is turned off to allow for users to decide if/when/how to apply (see the xml template above).
+ The test was chosen because there was clear "scalloping" (i.e. between-burst discontinuities in the unwrapped phase) in the existing GUNW ([link](https://search.asf.alaska.edu/#/?searchType=List%20Search&searchList=S1-GUNW-A-R-043-tops-20221002_20220908-151531-00041E_00027N-PP-e567-v2_0_5&resultsLoaded=true&granule=S1-GUNW-A-R-043-tops-20221002_20220908-151531-00041E_00027N-PP-e567-v2_0_5-amplitude)).


isce2_topsapp/topsapp_proc.py
```diff
@@ -50,6 +59,10 @@ def topsapp_processing(*,
     with open(TEMPLATE_DIR/'topsapp_template.xml', 'r') as file:
         template = Template(file.read())
 
+    geocode_list = GEOCODE_LIST_BASE.copy()
+    if estimate_ionosphere_delay:
+        geocode_list.append('merged/topophase.ion')
```
dbekaert, line 64: @cmarshak just to double check if this is the unwrapped topophase or the wrapped variant. 
Check if tehre is not a topophase.ion.unw or so

isce2_topsapp/topsapp_proc.py
```diff
@@ -25,6 +25,15 @@
 
 TEMPLATE_DIR = Path(__file__).parent/'templates'
 
+GEOCODE_LIST_BASE = ['merged/phsig.cor',
+                     'merged/filt_topophase.unw',
+                     'merged/los.rdr',
+                     'merged/topophase.flat',
```
dbekaert, line 31: Just double checking if this ever gets used in packaging into the product?

isce2_topsapp/topsapp_proc.py
```diff
@@ -25,6 +25,15 @@
 
 TEMPLATE_DIR = Path(__file__).parent/'templates'
 
+GEOCODE_LIST_BASE = ['merged/phsig.cor',
```
dbekaert, line 28: We should check if we use all of these? I would only geocode the ones that gets used in the product. which would be fitlered coherence and filtered unwrapped phase corrected for DEM and topophase.

cmarshak: These are taken from here: https://github.com/aria-jpl/ariamh/blob/develop/interferogram/sentinel/topsApp.xml.tmpl#L32

isce2_topsapp/topsapp_proc.py
cmarshak: Let's look at the products together. I was going to just include this raster in the product.

cmarshak:
Going to update example based on continued scalloping of the example initially given. More discussion can be described: https://github.com/isce-framework/isce2/issues/620

```
isce2_topsapp --reference-scenes S1B_IW_SLC__1SDV_20171117T145926_20171117T145953_008323_00EBAB_AFB8 \
              --secondary-scenes S1A_IW_SLC__1SDV_20171111T150004_20171111T150032_019219_0208AF_EE89 \
              --estimate-ionosphere-delay True \
              --do-esd True \
              --esd-coherence-threshold .5 \
              > topsapp_img.out 2> topsapp_img.err
```

dbekaert:
There are changes needed.

dbekaert:
@ehavazli Can you have a look into this as well?

ehavazli:
> **to clarify**
> * Its unclear what the figure refers. Would be good to plot: 1) unwrapped layer from the product (this should not have the correction applied), 2) the ionospheric layer from the product, 3) a difference plot between both.  My recollection on this example was slightly different. Likely it might be the  filtering and  90 m product that smears things out a bit.

**1. Unwrapped phase layer from the product**
Screen Shot 2022-11-16 at 11 59 06 AM

**2. Ionopshere layer from the product**
Screen Shot 2022-11-16 at 11 58 54 AM

**3. Difference between ionosphere and unwrapped phase**
Screen Shot 2022-11-16 at 11 59 17 AM


The first and second plots are masked where the values are 0.

I would like to point out that the ionosphere layer is larger than the unwrapped phase and therefore would result in a larger image than the unwrapped phase if **NOT** masked before applying any the correction.

PR #75, opened 2022-11-15
Raider Tropo Correction

cmarshak:
This is brings in new [raider](https://github.com/dbekaert/RAiDER/) developments into the GUNW for tropospheric delay computation. This is still a draft and is branched from #74 so that should be reviewed and merged first.

To run raider successfully, I had to update the environment to `>=3.10`. Using now both `isce2` and `isce3`.


cmarshak:
@dbekaert - confirming that we want to perform tropospheric correction on reference SLC only. Or am I missing something important for time-series analysis here.

dbekaert:
1) So we compute the delay for reference and we compute one for secondary.
2) We compute the differential delay just as we do for the insar pairing.
3) we convert it from delays in m unites to 2-way path delay so you are missing a lambda/(4 pi) factor

cmarshak:
@jhkennedy 

1. You suggested offline that we could have 2 docker containers within a plugin is that suggested? Or should we have separate plugin. Your input would be greatly appreciated. There are some considerations:
  - Hitting the ERA5 API takes a significant amount of time and is the biggest bottleneck for processing
  - If we need to separate the plugins, there is some accounting to get insar pair metadata e.g. acq datetimes, extent, etc.
  - Packaging - will literally be a few lines in `xarray` - assuming there is no more layers beyond this, I am ok with the SLC portion being disjointed, but might be good to have it cleaner for burst portion
2. Integration test - I can do this - but just wanted to get your thoughts

isce2_topsapp/tropo_proc.py
```diff
@@ -0,0 +1,98 @@
...
+                                                    yml_name='secondary_raider'
+                                                    )
+
+    ds_ref = xr.open_dataset(ref_delay_cube_path, engine='rasterio')
+    ds_sec = xr.open_dataset(sec_delay_cube_path, engine='rasterio')
+
+    # Wavelenth of S1 in meters according to
+    # https://sentinels.copernicus.eu/web/sentinel/technical-guides/sentinel-1-sar/sar-instrument
+    lambda_s1 = .055465763
+    ds_delay = (ds_ref - ds_sec) * (lambda_s1) / (4 * np.pi)
```
cmarshak, line 95: @dbekaert - how does this look?

Note this is doing the differential delay at all levels/variables in the group of the cube (i.e. both `hydro` and `wet`).

dbekaert: LGTM @cmarshak .

We could add a bit more complexity. Raider does not do temporal interpolation and i missed to mention that before.
1)  Raider currently rounds to closed available model UTC epic (i.e.  depends on the model) when you provide a satellite UTC time. We could do 4 raider computations to compute the allowed UTC before and after the satellite UTC. 
2) We do the temporal interpolation in time with linear weights for reference  and then for secondary dates.
3) you then compute the interferometric delay like you do above. Although, we should extract wavelength on the fly from within product meta-data so this becomes suitable for putting also in ALOS workflow etc.

Once you have a prototype would be good to do a quick test to the data to see if what is done is correctly. Take a small temporal baseline GUNW (e.g. 6 days) that does _not_ cover a large event like  EQ. 
- Plot A. A = compute unwrapped phase minus mean of the unwrapped phase 
- Plot B on same colorbar as A. B = compute 2D tropospheric correction minus mean of the tropospheric phase. Note to compute the troposheric phase as the sum of the hydrostatic and wet 3D cube intersected with the DEM. The latter should be simple scipy and should exist in ARIA tools and raider already (@sssangha or @jlmaurer)
- plot A-B on same colorbar as A.

Issue #79, opened 2022-12-06
Invalid datetime values in delivery json metadata in v0.2.0


asjohnston-asf:
GUNW products generated using DockerizedTopsApp v0.2.0 have invalid `sensing_start` and `sensing_stop` values in their `.json` metadata files. This causes these products to be rejected when attempting to ingest them into the ASF archive.

For example, from https://hyp3-a19-jpl-contentbucket-1wfnatpznlg8b.s3.us-west-2.amazonaws.com/819086fb-3fb1-4294-8cb2-c1effb453f52/S1-GUNW-D-R-087-tops-20221204_20221122-161627-00157W_00020N-PP-37a3-v2_0_6.json :

```
"sensing_start": "2022-12-04T16:16:12.000ZZ",
"sensing_stop": "2022-12-04T16:16:42.000ZZ",
```

This is due to a change in [asf_search v5.0.0](https://github.com/asfadmin/Discovery-asf_search/blob/master/CHANGELOG.md#500):
> Timestamps in geojson fields now include an explicit Z time zone indicator.

The new build of DockerizedTopsApp pulled in the new version of asf_search, but also still appends an additional `Z` to the timestamps  at https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/main/isce2_topsapp/delivery_prep.py#L134

cmarshak:
Thank you for basically solving this!

PR #80, opened 2022-12-06
Update Start/Stop Time for CMR Ingest Json

cmarshak:
Solves #79, which identifies reason for ingest pipeline breaking. There are json schema validation in the tests. Manually updated these.


isce2_topsapp/delivery_prep.py
```diff
@@ -131,8 +131,8 @@ def format_metadata(nc_path: Path,
     metadata.update({"ogr_bbox": ogr_bbox,
                      "reference_scenes": all_metadata['reference_scenes'],
                      "secondary_scenes": all_metadata['secondary_scenes'],
-                     "sensing_start": f'{ref_props["startTime"]}Z',
-                     "sensing_stop": f'{ref_props["stopTime"]}Z',
+                     "sensing_start": f'{ref_props["startTime"]}',
+                     "sensing_stop": f'{ref_props["stopTime"]}',
```
asjohnston-asf, line 135: This may no longer require an f-string, it might just be `"sensing_start": ref_props["startTime"],` (assuming that attribute is already a string and not a datetime object)

jhkennedy (approved):

PR #88, opened 2023-01-19
Region of interest

cmarshak:
Adds arguments `--region-of-interest` as `xmin ymin xmax ymax` and `--frame-id` as integer. This is branched of the `ionosphere` branch so that should be merged first (mainly on account of github actions fixed there).


cmarshak:
Once the tests have passed, I am going to merge this. This will provide better testing capabilities for raider and ionosphere.

Lots more tests have to be included within this repo (via pytest), but again, can't hold this up. Rather have things break so we can continue to debug the various pieces of the new processing workflow.

PR #91, opened 2023-01-26
Solid Earth Tides

cmarshak:
Computes solid earth tides via [pysolid](https://github.com/insarlab/PySolid) and this [example](https://github.com/ACCESS-Cloud-Based-InSAR/SET_GUNW) by @sssangha.

Basic notebook to demonstrate how to use and plots.


isce2_topsapp/__main__.py
```diff
@@ -95,6 +97,7 @@ def gunw_slc():
                         help='xmin ymin xmax ymax in epgs:4326', required=False)
     parser.add_argument('--estimate-ionosphere-delay', type=bool, default=False)
     parser.add_argument('--frame-id', type=int, default=-1)
+    parser.add_argument('--compute-solid-earth-tide', type=bool, default=False)
```
jhkennedy, line 100: The `type=bool` here won't work as expected; see https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/pull/95

cmarshak:
Here is a sample product with the Solid Earth Tide included: [link](https://gunw-development-testing.s3.us-west-2.amazonaws.com/solid-earth-tide/S1-GUNW-A-R-064-tops-20210723_20210711-015000-00119W_00033N-PP-6267-v2_0_6.nc).

cmarshak:
Putting this here for reference: https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/discussions/84

dbekaert:
@sssangha @cmarshak 
1) Please make proper use of longitude and latitude instead of x and y. We want the product to have a consistent feel to users across different groups. All for the GUNW is in WGS84 so lon lat should be fine to use.
2) Update the fill value not to be a nan; consistent feel with our other variable groups.
3) I note the use of double. Please justify the impact on accuracy versus float.
4) In a separate issue tickets or PR to make it documented: 
-  confirm correct use of sign when propagated up to mintpy
- Confirm stich-ability 
- Confirm for a ts stack it gives same results as if you use mintpy internal .

cmarshak:
>Please make proper use of longitude and latitude instead of x and y. We want the product to have a consistent feel to users across different groups. All for the GUNW is in WGS84 so lon lat should be fine to use.

Done

> Update the fill value not to be a nan; consistent feel with our other variable groups.

Done

> I note the use of double. Please justify the impact on accuracy versus float.

Sim uses the imaging geometry dataset (coordinates and SAR look angles) to generate the SET layer. It is cleanest to literally copy the format of this dataset to this layer (not to mention ensuring consistency with this similar layer). We would have to redo the code if we wanted float within the metadata (see [this](https://stackoverflow.com/a/18012628)). Honestly, I do not think there is any issue with accuracy as when `xarray` reads it is seen as the array is seen as `float64` type.

Here is an example of the updates you requested on a 2.0.4 dataset: [link](https://gunw-development-testing.s3.us-west-2.amazonaws.com/solid-earth-tide/S1-GUNW-A-R-064-tops-20210723_20210711-015001-35393N_33512N-PP-6267-v2_0_4.nc)

sssangha:
@cmarshak following our convo last week, I have pushed some updates to the dependencies in the env file. @mgovorcin and myself still need to work on the interfacing of ARIA-tools and mintpy to run formal end-to-end tests, but I'm hoping in the meantime to get an independent hack validation to verify it's consistency with the internal mintpy workflow.

Also the [SET hyp3 integration PR](https://github.com/ASFHyP3/hyp3/pull/1461) is ready, which follows same template as the PR which introduced the [ionosphere estimate option](https://github.com/ASFHyP3/hyp3/pull/1440), where by default the new option is set to default until we do more testing.

dbekaert:
@cmarshak Please inspect the layer, you will see that the geometry meta-date are all float. The only double we have is for lon lat and heights.

- For the example provided, confirm lon lat changes. 
- Fill value is still Nan

cmarshak:
> For the example provided, confirm lon lat changes.

Screenshot:
image

Code: https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/set/isce2_topsapp/solid_earth_tides.py#L140-L142

> Fill value is still Nan

Sorry - misread - thought you wanted it to be nan. Looks like several rasters have nodata as 0 (except connectedComponents which is -1). I don't know what this rasters range is for the `solidEarthTide` - if this raster's pixels can potentially includes 0, then we should use something else. Just say what you want and we can get the correct nodata.

dbekaert:
It very unlikely that it will be exactly 0. @sssangha thoughts?

cmarshak:
+ I made the nodata `0` as requested and updated the test to reflect that. We can change it back to `np.nan` if Sim feels otherwise.
+ Also, I was able to use `astype` to update the type of the group to `float32` in `xarray`, which now shows as `float` in panoply. Hope this keeps the product more consistent as desired.

cmarshak:
Here is the example [link](https://gunw-development-testing.s3.us-west-2.amazonaws.com/solid-earth-tide/S1-GUNW-A-R-064-tops-20210723_20210711-015001-35393N_33512N-PP-6267-v2_0_4.nc)

sssangha:
@dbekaert I agree unlikely to matter, but fyi looks like mintpy does actually set the `no_data_value` to `nan` in their SET stacks:
>   NO_DATA_VALUE                                   none

dbekaert:
There is a difference between what the product does and how ARIA-tools
parses it to mintPY.

dbekaert:
As long as we are GDAL, QGIS, ARCGIS compatible thats the main driver.
The mintpy interface should be handled by ARIA-tools and should not drive
the no-date for gunw.

cmarshak:
@dbekaert - once you approve - it will automatically be merged (or anyone else outside of Sim and me).

sssangha:
Apologies again for the confusion @cmarshak , after discussion with @dbekaert we agreed it is best to be internally consistent with other variable groups. We could handle nan assignment with mintpy -- if necessary -- through the prep_aria pipeline.

I've re-reverted your changes to reflect a nodata assignment of 0.

dbekaert (approved):
Lgtm

PR #92, opened 2023-01-30
Use just the frame-id for CLI

cmarshak:
Per @asjohnston-asf suggestion, utilize just the frame-id to submit jobs.

I am reading directly from an s3 bucket. We could also add the dependency from [s1-frame-enumerator](https://github.com/ACCESS-Cloud-Based-InSAR/s1-frame-enumerator) and read it from that.

### DIFF
--- a/isce2_topsapp/localize_slc.py
+++ b/isce2_topsapp/localize_slc.py
@@ -2,6 +2,7 @@ import netrc
 from concurrent.futures import ThreadPoolExecutor
 
 import asf_search as asf
+import geopandas as gpd
 from shapely.geometry import GeometryCollection, box, shape
 from shapely.ops import unary_union
 from tqdm import tqdm
@@ -32,12 +33,16 @@ def get_session():
 
 def check_geometry(reference_obs: list,
                    secondary_obs: list,
-                   region_of_interest: list) -> GeometryCollection:
+                   frame_id: int = -1) -> GeometryCollection:
     reference_geos = [shape(r.geojson()['geometry']) for r in reference_obs]
     secondary_geos = [shape(r.geojson()['geometry']) for r in secondary_obs]
 
     reference_geo = unary_union(reference_geos)
     secondary_geo = unary_union(secondary_geos)
+
+    # if they are not Polygons they are multipolygons and not valid
+    connected_ref = (reference_geo.geom_type == 'Polygon')
+    connected_sec = (secondary_geo.geom_type == 'Polygon')
 
     # Two geometries must intersect for their to be an interferogram
     intersection_geo = secondary_geo.intersection(reference_geo)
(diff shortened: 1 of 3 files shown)


isce2_topsapp/localize_slc.py
```diff
@@ -32,39 +33,43 @@ def get_session():
...
                            'is empty')
 
     # Update the area of interest based on user specification
-    if region_of_interest is not None:
-        region_of_interest_geo = box(*region_of_interest)
-        if not region_of_interest_geo.intersects(intersection_geo):
-            raise RuntimeError('Region of interest does not overlap with IFG '
-                               'area (ref and sec overlap)')
+    if frame_id != -1:
+        df_frames = gpd.read_file('s3://s1-gunw-frames/s1_frames.geojson',
```
jhkennedy, line 59: This file is only accessible using JPL A19 AWS credentials and won't be accessible from other deployments, or to developers locally (without effort). If you're going to keep it in an S3 bucket, it'd be best to make it a public bucket. 

However, since this file is only 72 mb, and compresses down to 22 mb with zip, I think you'd be best off just including it in the repo. That way, it's guaranteed to work for local developers, in the container, and in any deployment. Also, that way changes to the frames are explicitly  captured in PRs/git history, and documented in the changelog.

jhkennedy: Adding the enumerator as a dependency and reading from that would also work.

cmarshak: I am going to add the enumerator as a dependency. There are issues with the frame size/overlap as it takes too long. Will get clarity soon from @dbekaert.

cmarshak:
@jhkennedy and @asjohnston-asf 

This is actually ready for review.

Notes:

1. Included frame data as zip file in this repository. Will be easiest currently because the enumerator is really 3.10+ and there are no tests yet for this.
2. Included a bunch of integration tests that can be easily ported to unit tests once there is a bit more time to use the examples Andrew shared with me earlier this week. Do a bunch of checks to make sure we are submitting correct pairs
3. Had to pin ISCE2 to version 2.6.1 (the previous version) with numpy/scipy pinned too until some build issues are sorted out. We will likely want to figure out a better way to increment version numbers for ISCE2 since the errors are pretty cryptic. Thanks, @sssangha.
4. Most importantly, frame id CLI is there. See the README for examples. We are using 8 bursts across all 3 swaths with a guarantee of 1 burst overlap (usually more because we are using a bbox to specify processing area in ISCE2).

Here is the GUNW for the first example [[link](https://gunw-development-testing.s3.us-west-2.amazonaws.com/frame-testing/S1-GUNW-D-R-144-tops-20230125_20221220-140020-00121W_00036N-PP-28c2-v2_0_6.nc)]. For additional context, here are some geojsons of the [frame 22438](https://gunw-development-testing.s3.us-west-2.amazonaws.com/frame-testing/22438_frame.geojson) and its bounding [box](https://gunw-development-testing.s3.us-west-2.amazonaws.com/frame-testing/22438_box.geojson).

cmarshak:
Here is the same data for an overlapping frame

GUNW: https://gunw-development-testing.s3.us-west-2.amazonaws.com/frame-testing/S1-GUNW-D-R-144-tops-20230125_20221220-140045-00121W_00035N-PP-4569-v2_0_6.nc
Frame 22439: https://gunw-development-testing.s3.us-west-2.amazonaws.com/frame-testing/22439_frame.geojson
Bbox: https://gunw-development-testing.s3.us-west-2.amazonaws.com/frame-testing/22439_box.geojson

jhkennedy (approved):
I think this looks good now; but I wasn't able to open the NetCDF you posed because both QGIS and ncview give me an error like:
```
ncview: can't recognize format of input file S1-GUNW-D-R-144-tops-20230125_20221220-140045-00121W_00035N-PP-4569-v2_0_6.nc
```

jhkennedy:
That is, in terms of testing the hyp3 integration, this looks good to me. But the actual product produced likely needs work.

cmarshak:
The netcdf thing doesn't suprise me. I get similar errors with xarray. We have to use `h5py` to open GUNWs.

I can view the product in QGIS (or at least certain rasters) using the QGIS LTR (long-term release). The newest versions of QGIS always fail.

image
image

PR #95, opened 2023-01-31
fix boolean arguments in GUNW entrypoints

jhkennedy:
As documented in the README, these arguments expect strings like `'True`'/`'False'`:
* `--estimate-ionosphere-delay`
* `--do-esd`

However, the argument parser uses a `bool` type for these arguments, which evaluates *any* non-empty string as `True`:

```python
>>> bool('true')
True
>>> bool('True')
True
>>> bool('False')
True
>>> bool('false')
True
>>> bool('foobar')
True
>>> bool('')
False
```

This adds a comparison function that can be used as a `type` for the argument parser and only returns true if the string `true` (any capitalization) is provided. 

```python
>>> string_is_true('true')
True
>>> string_is_true('True')
True
>>> string_is_true('False')
False
>>> string_is_true('false')
False
>>> bool('foobar')
False
>>> bool('')
False
```

in-line we the `False` default.

---

Note: Since none of these arguments have been released, I've labeled this as `bumpless`.

### DIFF
--- a/isce2_topsapp/__main__.py
+++ b/isce2_topsapp/__main__.py
@@ -82,6 +82,9 @@ def ensure_earthdata_credentials(username: Optional[str] = None, password: Optio
         )
 
 
+def string_is_true(s: str) -> bool:
+    return s.lower() == 'true'
+
 def gunw_slc():
     parser = ArgumentParser()
     parser.add_argument('--username')
@@ -93,10 +96,10 @@ def gunw_slc():
     parser.add_argument('--secondary-scenes', type=str.split, nargs='+', required=True)
     parser.add_argument('--region-of-interest', type=float, nargs=4, default=None,
                         help='xmin ymin xmax ymax in epgs:4326', required=False)
-    parser.add_argument('--estimate-ionosphere-delay', type=bool, default=False)
+    parser.add_argument('--estimate-ionosphere-delay', type=string_is_true, default=False)
     parser.add_argument('--frame-id', type=int, default=-1)
-    parser.add_argument('--do-esd', type=bool, default=False)
-    parser.add_argument('--esd-coherence-threshold', type=float, default=-1)
+    parser.add_argument('--do-esd', type=string_is_true, default=False)
+    parser.add_argument('--esd-coherence-threshold', type=float, default=-1.)
     args = parser.parse_args()
 
     do_esd_arg = (args.esd_coherence_threshold != -1) == args.do_esd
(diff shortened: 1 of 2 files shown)


isce2_topsapp/__main__.py
```diff
@@ -82,6 +82,10 @@ def ensure_earthdata_credentials(username: Optional[str] = None, password: Optio
         )
 
 
+def string_is_true(s: str) -> bool:
+    return s.lower() == 'true'
+
```
cmarshak, line 87: Should we add a check that `s.lower()` is in `['true', 'false']` so there are no unexpected results?

cmarshak:
I had no idea - clearly not versed in argparse and these common pitfalls. I super appreciate you catching this because this would be pretty bad if we were wondering why ionosphere was always being applied.

Thank you, Joe!

cmarshak (approved):
Looks great. Thank you.

isce2_topsapp/__main__.py
jhkennedy: @cmarshak you could add something like:
```python
if s.lower() is not in ('true', 'false'):
    raise argparse.ArgumentError('Only the strings `true` or `false` (any capitalization) may be provided.')
```
if you wanted it to crash if a nonsensical string was provided. We don't for a strictly HyP3 API, but for user-run stuff it's probably worth while.

jhkennedy: Should add a test or two at that point though

jhkennedy: :thinking: probably the right thing to do, so I'll do that

cmarshak:
Thanks so much. Looks great.

We will reuse this for SET (@sssangha). Super grateful you caught this.

cmarshak:
Happy to write some simple tests later using your nice snippet above :)

jhkennedy:
> Happy to write some simple tests later using your nice snippet above :)

I did add the tests ;-)

cmarshak:
🙇

tests/test_main.py
```diff
@@ -71,3 +71,19 @@ def test_main_check_earthdata_credentials_env_variables(tmp_path, monkeypatch):
...
+    assert true_false_string_argument('false') is False
+    assert true_false_string_argument('False') is False
+    assert true_false_string_argument('FALSE') is False
+
+    with pytest.raises(ValueError):
+        true_false_string_argument('foo')
+    with pytest.raises(ValueError):
+        true_false_string_argument('bar')
+    with pytest.raises(ValueError):
+        true_false_string_argument('')
```
cmarshak, line 89: Noice

cmarshak (approved):

PR #98, opened 2023-02-03
Illustrate how to add additional 2d layers

cmarshak:
Shows how to add a "dummy" ionosphere layer to the GUNW product. Serves as a reminder how to do this and test additional layers that we may wish to add.

It's an "integration" test. It would be best to use the notebook diff tool to view it. Again, should improve these generally to run faster, but these are serving both to ensure correctness (in case they get edited) and how-to in terms of adding these layers.


gracebato (approved):
Looks good to me.

PR #99, opened 2023-02-07
Update actions to use tags instead of `main`

jhkennedy:
Currently, the test build actions are failing because of an invalid workflow file:
image

ASFHyP3/Actions switch to micromamba in v0.7.0 and no longer needs to know that conda environment name, so this parameter is no longer valid. 

This PR:
* Drops `conda_env_name` from all workflows
* pins the ASFHyP3/Actions versions
* Adds dependabot to open PRs to update action versions

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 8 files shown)


jhkennedy:
@cmarshak pytest seems to be failing with auth errors

jtherrmann (approved):

PR #103, opened 2023-02-14
Update Frame Geojson

cmarshak:
Uses frame footprints (frames and all overlapping bursts) to ensure that enough DEM is requested. The frames used to determine SLCs are slightly smaller. This is discussed [here](https://github.com/ACCESS-Cloud-Based-InSAR/s1-frame-generation/blob/main/2%20Latitude%20Aligned%20Frames.ipynb).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,4 +105,4 @@ tests/out/*.ipynb
 tests/**/*.json
 
 # Use geojson zip file for data
-!isce2_topsapp/data/gunw_frame_footprints.geojson.zip
+!isce2_topsapp/data/s1_gunw_frame_footprints.geojson.zip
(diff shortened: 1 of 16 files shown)


cmarshak:
So,

1. `s1_frames_latitude_aligned.geojson.zip` determine the `processing_extent` for ISCE2 - this is the bbox to be submitted in the topsApp XML. This will be slightly smaller than the final GUNW which will include all the bursts that overlap with this.
2. `s1_gunw_frame_footprints.geojson.zip` are the extent determined by taking all intersecting bursts with the above.

The latter is used to submit extents for getting the DEM since for geocoding the entire GUNW area must be covered.

This image summarizes it:
image

The polygons with transparency are the latitude aligned frames, the solid color outlines are the gunw extents.

cmarshak:
Also, updated `intersection_geometry` to `gunw_geometry` or (`ifg_geometry`) because the latter is more descriptive of the variable and metadata.

sssangha (approved):
LGTM!

PR #108, opened 2023-02-23
Update Version Attribute

### LINKED ISSUE #107: Version in metadata not properly defined.
**Describe the bug**
GUNW metadata version written out as integer "1" as opposed to expected "1c".

**Additional context**
Suspect it's related to the use of the h5py module. I.e. the following approach does not update the metadata field as expected as the letter "c" is stripped:
```
with h5py.File(nc_path, mode='a') as file:
    file.attrs.modify('version', '1c')
```

But using the netCDF module instead writes out the string as expected:
from netCDF4 import Dataset
```
with Dataset(nc_path, mode='a') as ds:
    ds.version = '1c'
```

We need to track exactly where this field is updated/set in the code.

cmarshak:
Resolves #107. Ensures `1b` version attribute is changed to `1c`.

Not really sure about internals of different netcdf drivers and the implications. What caused the issue was that version attribute was read as `numpy.bytes_` with fixed length by `h5py`, whereas for `netCDF4` the same attribute was read as a string and length is variable. The fix using the former driver was to delete the attribute and overwrite it. Anyways, GUNWs are a decade in the making, so might be helpful why we are using so many different drivers.

Have a small notebook to illustrate the curiosity: https://gist.github.com/cmarshak/1822cd22fea0cc634cf026bfccbb4aa7

Added a test of the version update to ensure version update to `1c` going forward.

Also fixed the control flow, which was wrong because the version attribute needs to be updated if SET or ionosphere is added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.2.3]
-
-### Added
-* Added support to compute and embed solid earth tide correction layers into GUNW products (see PR #91)
-
 ## [0.2.2]
 
 ### Added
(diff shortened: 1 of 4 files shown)


sssangha (approved):
LGTM

PR #109, opened 2023-02-27
Fix variable name error

cmarshak:
I recently renamed `intersection_geo` to `ifg_geo` since with frames dictating extents, the latter makes more sense. Not sure why this variable change (or lack of it) wasn't caught during tests/linting - but glad we have these integration tests.


gracebato:
Looks good to me.

gracebato (approved):

PR #110, opened 2023-02-28
Update Solid Earth Tide Correction to reference and secondary groups

cmarshak:
Splits reference and secondary solid earth tide corrections into separate groups:

specifically

```
/science/grids/corrections/external/tides/solidEarth/reference
/science/grids/corrections/external/tides/solidEarth/secondary
```


sssangha (approved):
LGTM

PR #111, opened 2023-02-28
Remove dummy path with h5py (fix pathname)

cmarshak:
The packaging puts a string placeholder. The path was incorrectly named.


sssangha (approved):
Good catch!

PR #112, opened 2023-03-02
Makes sure that the date and time in the GUNW ID are derived from center acquisition datetime

cmarshak:
A GUNW ID looks like this:

```
S1-GUNW-D-R-048-tops-20230106_20221214-235959-00090E_00040N-PP-c254-v2_0_6
```
The `20230106` is the reference date and `20221214` is the secondary date. The `235959` is the center time i.e. `23:59:59`.

Previously the dates in the GUNW id were derived from start time. The center time was indeed the center time of a pass. In other words, these two date and time tokens were derived from _different_ datetimes and could be problematic if assuming not - i.e. using the GUNW ID to extract the metadata. This brought up in Raider [here](https://github.com/dbekaert/RAiDER/issues/488). You can see the current GUNW ID would be `[S1-GUNW-D-R-048-tops-20230106_20221214-000000-00090E_00040N-PP-c254-v2_0_6.nc](https://hyp3-a19-jpl-contentbucket-1wfnatpznlg8b.s3.us-west-2.amazonaws.com/5225cd05-0869-4031-a697-15778a29635f/S1-GUNW-D-R-048-tops-20230106_20221214-000000-00090E_00040N-PP-c254-v2_0_6.nc)` which is confusing (and wrong) if you think the center time and the start time correspond to the same datetime, which would lead you to believe the GUNW was acquired 24 hours before it did! For more details, see the linked issue ticket.

Also cleaned up the code to make it more readable. Added a test which was hardest part because had to copy metadata (I did make up the `stopTimes` - just added 25 seconds).

### DIFF
--- a/isce2_topsapp/packaging.py
+++ b/isce2_topsapp/packaging.py
@@ -1,3 +1,4 @@
+import datetime
 import hashlib
 import json
 import os
@@ -6,7 +7,6 @@ from pathlib import Path
 from typing import Union
 
 import h5py
-import numpy as np
 from dateparser import parse
 
 import isce2_topsapp
@@ -64,18 +64,16 @@ def get_geo_str(extent: list) -> str:
     return f'{lon_str}_{lat_str}'
 
 
-def get_center_time(properties: list) -> str:
+def get_center_time(properties: list) -> datetime.datetime:
 
-    ref_start_times = [parse(props['startTime']) for props in properties]
-    ref_stop_times = [parse(props['stopTime']) for props in properties]
+    start_times = sorted([parse(props['startTime']) for props in properties])
+    stop_times = sorted([parse(props['stopTime']) for props in properties])
 
-    all_times = (ref_start_times + ref_stop_times)
-    N = len(all_times)
-    all_time_deltas = [all_times[k] - all_times[0] for k in range(N)]
+    start_time = start_times[0]
+    stop_time = stop_times[-1]
 
-    center_time = all_times[0] + np.mean(all_time_deltas)
-
-    return center_time.strftime('%H%M%S')
+    center_datetime = start_time + (stop_time - start_time) / 2
+    return center_datetime
 
 
 def get_gunw_id(reference_properties: list,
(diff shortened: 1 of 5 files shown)


isce2_topsapp/packaging.py
```diff
@@ -92,21 +90,22 @@ def get_gunw_id(reference_properties: list,
...
+
+    # Center Time sring
+    ref_center_time_str = ref_center_datetime.strftime('%H%M%S')
+
+    reference_date_str = ref_center_datetime.strftime('%Y%m%d')
+    secondary_date_str = sec_center_datetime.strftime('%Y%m%d')
 
     # date pair
-    date_pair = f'{reference_date}_{secondary_date}'
+    date_pair = f'{reference_date_str}_{secondary_date_str}'
```
bbuzz31, line 104: As long as this is what goes into the GUNW filename, then I think we're good!

bbuzz31 (commented):
LGTM pending comment confirmation

isce2_topsapp/packaging.py
```diff
@@ -92,21 +90,22 @@ def get_gunw_id(reference_properties: list,
...
-    # dates; remove dashes
-    reference_date = reference_properties[0]['startTime'].split('T')[0]
-    reference_date = reference_date.replace('-', '')
-    secondary_date = secondary_properties[0]['startTime'].split('T')[0]
-    secondary_date = secondary_date.replace('-', '')
+    # Center Datetimes
+    ref_center_datetime = get_center_time(reference_properties)
+    sec_center_datetime = get_center_time(secondary_properties)
+
+    # Center Time sring
```
dbekaert, line 97: typo

isce2_topsapp/packaging.py
cmarshak: Yes - look at the test and make sure!

The center datetimes of reference and secondary are now used to generate the GUNW id.

cmarshak: https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/e9dccb7b4b14326dacf0540cdd333a6f40f1a327/isce2_topsapp/packaging.py#L119-L136

dbekaert: looks good for the filename.

dbekaert (approved):
LGTM. Can you test an old region that should not have changed filename with the new one to confirm?

cmarshak:
Hey @dbekaert - the center time will likely be *slightly* different.

In original HySDS, the formula was
- `min_acq_time + (max_acq_time - min_acq_time) / 2

For our plugin (until now), it was:
- `all_acq_times[0] + mean([acq_time - all_acq_times[0] for acq_time in all_acq_times[1:]])`

I was trying to avoid sorting acq times - but the former is a lot clearer and easier to read (sorting is not a big time sink!). 

The main point which was present in this plugin and the previous HySDS (I believe) was to ensure that center time was the *same* as the `acq_date` in the GUNW id.

PR #114, opened 2023-03-16
Update production with fixes particularly related to ASF Metadata and DEM near dateline

cmarshak:
There is a lot in this production merge due to ongoing development.

The new features are by default turned off within our plugin (e.g. ionosphere is not performed). Moreover, the hyp3 api interface does not currently expose these parameters.

We may want to do an incremental merge of individual files in the plugin, but at this point, it won't hurt to merge since the only processing requests will be Aleutians (needs dateline fixes) and disasters (often needs custom corrections as it is).


jhkennedy:
@cmarshak if you want to also make  the options in `INSAR_ISCE_TEST` available in production, we'd only need to update:
https://github.com/ASFHyP3/hyp3/blob/develop/job_spec/INSAR_ISCE.yml
to match:
https://github.com/ASFHyP3/hyp3/blob/develop/job_spec/INSAR_ISCE_TEST.yml
except any of the `image_tag: test` lines would need to be removed.

cmarshak:
We may want to merge this for our production branch and not make the correction options available yet. We need some of the dateline transformations of metadata and DEM that were fixed.

Will ask @dbekaert.

cmarshak:
Again, don't need to expose parameters for corrections (yet) - they are by default turned off so we are good for production. Want to do some Aleutian processing. Thanks!

cmarshak:
@jhkennedy - can you confirm even if hyp3 does not have API entrypoints/parameters for the correction layers, we can merge the `dev` branch into production?

Thanks!

jhkennedy:
@cmarshak release failed because the changelog has a new PATCH version (0.2.1 -> 0.2.2) but this was tagged 'minor'

jhkennedy:
@cmarshak if you push the correct release tag, that'll re-trigger the release process:
```shell
git tag -a v0.2.2 -m "Release v0.2.2" cda63f5
```

jhkennedy:
@cmarshak it's okay to just merge #115 to bring main into develop, bu you can also do the fast forward merge yourself to prevent a merge commit:
```shell
git switch dev
git pull --ff-only
git merge --ff-only origin/main
git push
```

PR #116, opened 2023-03-21
Remove python 3.8 support

cmarshak:
It appears 3.9 builds fine. We need to add 3.10 soon!


sssangha (approved):
LGTM. Also was necessary for a stable build of ARIA-tools

PR #123, opened 2023-03-27
Iono workflow with masking

mgovorcin:
Resolves #85.

Adding option to use water mask for ionospheric correction:
       --      Ionospheric computation is done after topsApp run using ionoApp.xml
       --      Iono_processiong runs in 5 steps: 

             1. run ion step 'subband' (creation of lower and upper fine_interferograms for each burst)
             2. transform (water) mask from geo to radar coordinate space, and us it to mask burst interferograms
             3. run ion steps 'rawion' to 'esd' (computation and filtering of ionospheric correction)
             4. create merged/topophase.ion (extracted from runMergeBursts.py)
             5. run geocode for merged/topophase.ion

### DIFF
--- a/isce2_topsapp/iono_proc.py
+++ b/isce2_topsapp/iono_proc.py
@@ -9,315 +9,280 @@ import numpy as np
 from jinja2 import Template
 from tqdm import tqdm
 
-from isce.components import isceobj
 from numpy.typing import NDArray
+from isce.components import isceobj
+from isce.components.isceobj.TopsProc.runMergeBursts import (
+    interpolateDifferentNumberOfLooks,
+)
 
-try:
-    from osgeo import gdal
-except ModuleNotFoundError as exc:
-    raise ModuleNotFoundError('Python module gdal missing!') from exc
+from osgeo import gdal
 
-'''
-List of parameters for ionospheric correction:
-______________________________________________________________________
-property name="do ionosphere correction">False</property>
-<property name="apply ionosphere correction">False</property>
-<property name="consider burst properties in ionosphere computation">False</property>
-#choose from: ['subband', 'rawion', 'grd2ion', 'filt_gaussian', 'ionosphere_shift', 'ion2grd', 'esd']
-<property name="start ionosphere step">subband</property>
-<property name="end ionosphere step">esd</property>
+# List of parameters for ionospheric correction:
+# https://github.com/isce-framework/isce2/blob/8af43d2b9f291e2370e9112a3ead1449d0f2a386/examples/input_files/topsApp.xml#L182
 
-<property name="height of ionosphere layer in km">200.0</property>
-<property name="apply polynomial fit before filtering ionosphere phase">True</property>
-<property name="maximum window size for filtering ionosphere phase">200</property>
-<property name="minimum window size for filtering ionosphere phase">100</property>
-<property name="maximum window size for filtering ionosphere azimuth shift">150</property>
-<property name="minimum window size for filtering ionosphere azimuth shift">75</property>
-#0: no correction
-#1: use mean value of a burst
-#2: use full burst
-<property name="correct phase error caused by ionosphere azimuth shift">1</property>
+IONO_STEPS = ["subband", "rawion", "grd2ion", "filt_gaussian"]
 
-seperated islands or areas usually affect ionosphere estimation and it's better to mask them
-#out. check ion/ion_cal/raw_no_projection.ion for areas to be masked out.
-#The parameter is a 2-D list. Each element in the 2-D list is a four-element list: [firstLine,
-#lastLine, firstColumn, lastColumn], with line/column numbers starting with 1. If one of the
-#four elements is specified as -1, the program will use firstLine/lastLine/firstColumn/
-#lastColumn instead. For exmple, if you want to mask the following two areas out, you can
-#specify a 2-D list like:
-#[[100, 200, 100, 200],[1000, 1200, 500, 600]]
-<property name="areas masked out in ionospheric phase estimation">None</property>
+TEMPLATE_DIR = Path(__file__).parent / "templates"
 
-#better NOT try changing the following two parameters, since they are related
-#to the filtering parameters above
-<property name="total number of azimuth looks in the ionosphere processing">50</property>
-<property name="total number of range looks in the ionosphere processing">200</property>
-#the above numbers should be integer multiples of the below numbers
-<property name="number of azimuth looks at first stage for ionosphere phase unwrapping">10</property>
-<property name="number of range looks at first stage for ionosphere phase unwrapping">40</property>
-'''
-
-IONO_STEPS = ['subband',
-              'rawion',
-              'grd2ion',
-              'filt_gaussian',
-               # only do the following steps 
-               # when considering burst properties
-              'ionosphere_shift',
-              'ion2grd',
-              'esd']
-
-TEMPLATE_DIR = Path(__file__).parent/'templates'
-
-GEOCODE_LIST_BASE = ['merged/topophase.ion']
+GEOCODE_LIST_BASE = ["merged/topophase.ion"]
 
 
-def iono_processing(*,
-                    reference_slc_zips: list,
-                    secondary_slc_zips: list,
-                    orbit_directory: str,
-                    extent: list,
-                    dem_for_proc: str,
-                    dem_for_geoc: str,
-                    azimuth_looks: int = 7,
-                    range_looks: int = 19,
-                    swaths: list = None,
-                    do_esd: bool = False,
-                    esd_coherence_threshold: float = .7,
-                    mask_filename: str = None):
+def iono_processing(
+    *,
+    reference_slc_zips: list,
+    secondary_slc_zips: list,
+    orbit_directory: str,
+    extent: list,
+    dem_for_proc: str,
+    dem_for_geoc: str,
+    azimuth_looks: int = 7,
+    range_looks: int = 19,
+    swaths: list = None,
+    do_esd: bool = False,
+    esd_coherence_threshold: float = 0.7,
+    mask_filename: str = None,
+):
     swaths = swaths or [1, 2, 3]
     # for [ymin, ymax, xmin, xmax]
     extent_isce = [extent[k] for k in [1, 3, 0, 2]]
 
     # Update PATH with ISCE2 applications
-    isce_application_path = Path(f'{site.getsitepackages()[0]}'
-                                    '/isce/applications/')
-    os.environ['PATH'] += (':' + str(isce_application_path))
+    isce_app_path = Path(f"{site.getsitepackages()[0]}" "/isce/applications/")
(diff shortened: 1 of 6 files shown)


dbekaert:
Does ESD need to be disabled when using iono?

mgovorcin:
No, esd is only used when iono correction is done on burst level. By default is skipped.

isce2_topsapp/iono_proc.py
```diff
@@ -0,0 +1,383 @@
...
+from jinja2 import Template
+from tqdm import tqdm
+
+from isce.components import isceobj
+from numpy.typing import NDArray
+
+try:
+    from osgeo import gdal
+except ModuleNotFoundError as exc:
+    raise ModuleNotFoundError('Python module gdal missing!') from exc
```
cmarshak, line 18: No need to use this - gdal is in our environment. Needs to be.

isce2_topsapp/iono_proc.py
```diff
@@ -0,0 +1,383 @@
...
+<property name="areas masked out in ionospheric phase estimation">None</property>
+
+#better NOT try changing the following two parameters, since they are related
+#to the filtering parameters above
+<property name="total number of azimuth looks in the ionosphere processing">50</property>
+<property name="total number of range looks in the ionosphere processing">200</property>
+#the above numbers should be integer multiples of the below numbers
+<property name="number of azimuth looks at first stage for ionosphere phase unwrapping">10</property>
+<property name="number of range looks at first stage for ionosphere phase unwrapping">40</property>
+'''
```
cmarshak, line 58: No need to copy this from ISCE2 - just link.

isce2_topsapp/iono_proc.py
```diff
@@ -0,0 +1,383 @@
...
+                    'range_looks': range_looks,
+                    'swaths': swaths,
+                    'geocode_list': GEOCODE_LIST_BASE
+                     }
+
+    with open(TEMPLATE_DIR/'topsapp_iono_template.xml', 'r') as file:
+        template = Template(file.read())
+
+    # Step-1 Run ionospheric correction step subband
+    iono_xml = template.render(**iono_kwargs,
```
cmarshak, line 126: For clarity and for error handling, if possible, break out each step into its own function. That way if errors occur, we can individually test functions.

isce2_topsapp/iono_proc.py
```diff
@@ -0,0 +1,383 @@
...
+        image = isceobj.createImage()
+        image.setDataType('FLOAT')
+        image.setFilename(ionFiltOut)
+        image.extraFilename = ionFiltOut + '.vrt'
+        image.setWidth(img.width)
+        image.setLength(img.length)
+        # image.setAccessMode('read')
+        # image.createImage()
+        image.renderHdr()
+        # image.finalizeImage()
```
cmarshak, line 273: Why are these commented out? If you think someone may want to comment them back for debugging, please explain.

isce2_topsapp/iono_proc.py
```diff
@@ -0,0 +1,383 @@
...
+    lon_ds = gdal.Open(lonFilename + '.vrt')
+    lat_ds = gdal.Open(latFilename + '.vrt')
+
+    # Get lon and lat arrays
+    lons = lon_ds.ReadAsArray()
+    lats = lat_ds.ReadAsArray()
+
+    # Translate mask geo corrdinate to radar
+    lineIndex = np.int32((lats - mask_ds.GetGeoTransform()[3]) / mask_ds.GetGeoTransform()[5] + 0.5)
+    sampleIndex = np.int32((lons - mask_ds.GetGeoTransform()[0]) / mask_ds.GetGeoTransform()[1] + 0.5)
```
cmarshak, line 312: Can you please explain `0.5` in a comment?

isce2_topsapp/iono_proc.py
```diff
@@ -0,0 +1,383 @@
...
+        image.setFilename(outputFilename)
+        image.extraFilename = outputFilename + '.vrt'
+        image.setWidth(mask_radarcoord.shape[1])
+        image.setLength(mask_radarcoord.shape[0])
+        image.renderHdr()
+
+    return mask_radarcoord
+
+
+def mask_interferogram(ifgFilename: Union[str, Path],
```
cmarshak, line 340: you could potentially also name this `mask_raster_in_rdr` or whatever you like since again this could be repurposed.

isce2_topsapp/iono_proc.py
```diff
@@ -0,0 +1,383 @@
...
+                       maskArray: NDArray,
+                       outFilename: Union[str, Path] = None) -> None:
+    
+    '''
+    This routine uses mask np.array to mask wrapped interferogram
+    ifgFilename : str
+            path to georeferenced mask raster (must have vrt extension)
+    maskArray : np.array
+            mask 2d array (1: mask values, 0; no-mask values))
```
cmarshak, line 349: In rdr coordinates right?

isce2_topsapp/iono_proc.py
```diff
@@ -0,0 +1,383 @@
...
+    outFilename : str
+            path where to save masked interferogram (default: None = overwrite existing)
+    '''
+
+    # Read interferogram
+    int_ds = gdal.Open(ifgFilename + '.vrt')
+    int_array = int_ds.ReadAsArray()
```
cmarshak, line 356: I would make sure the dimensions are the same as well and if not, raise ValueError.

isce2_topsapp/localize_mask.py
```diff
@@ -0,0 +1,32 @@
...
+    
+    if water_name == 'SWDB':
+        import isce
+        from isceobj.Alos2Proc.runDownloadDem import download_wbd
+        #Download SRTM-SWDB water mask
+        mask_filename = download_wbd(extent_buffered[1],
+                                     extent_buffered[3], 
+                                     extent_buffered[0], 
+                                     extent_buffered[2])
+
```
cmarshak, line 26: Nice!

isce2_topsapp/iono_proc.py
```diff
@@ -0,0 +1,383 @@
...
+              burstIndex, box, os.path.join(mergedir, mergedIonname+suffix), virtual=virtual, validOnly=True)
+        multilook(os.path.join(mergedir, mergedIonname+suffix),
+                      outname = os.path.join(mergedir, mergedIonname),
+                      alks = self.numberAzimuthLooks, rlks=self.numberRangeLooks)
+        '''
+    else:
+        ionFilt = 'ion/ion_cal/filt.ion'
+        img = isceobj.createImage()
+        img.load(ionFilt+'.xml')
+        ionFiltImage = (np.fromfile(ionFilt, dtype=np.float32).reshape(img.length*2, img.width))[1:img.length*2:2, :]
```
cmarshak, line 252: Could you please add a comment as to why this indexing is happening and/or link to the code in ISCE2 you are using?

isce2_topsapp/iono_proc.py
```diff
@@ -0,0 +1,383 @@
...
+    if result.returncode != 0:
+        raise ValueError('TopsApp failed at step: ion-subband')
+
+    # Step-2 Mask upper and lower band burst wrapped interferograms
+    # path to burst ifgs: ion/lower|upper/fine_interferogram/IW{1,2,3}/burst_{##}.int
+    # path to burst geometry: geom_reference/IW{1,2,3}/lat|lon_{##}.rdr
+
+    if mask_filename:
+        # Project mask to burst image coordinate space
+        workdir = Path('.').absolute()
```
cmarshak, line 147: Use `resolve` per [this](https://stackoverflow.com/a/44569249).

isce2_topsapp/iono_proc.py
```diff
@@ -0,0 +1,383 @@
...
+
+        get_swath = lambda x: x.split('/')[-2]
+        get_burst = lambda x: x.split('/')[-1].split('_')[-1].split('.')[0]
+
+        # Get all geometry files
+        geom_files = list(workdir.glob('geom_reference/IW*/lat*.rdr'))
+
+        # NOTE: THis can be easily parallized, skipped for now
+        with tqdm(total=len(geom_files)) as pbar:
+            for lat in geom_files:
```
cmarshak, line 157: Why not just use something like:

```
from tqdm import tqdm

for lat in tqdm(geom_files):
```
Simplifies a little.

isce2_topsapp/iono_proc.py
mgovorcin: I took these lines from runMergeBursts.py, not sure the rational...I deleted the commented line and added link to the reference script

isce2_topsapp/iono_proc.py
mgovorcin: I moved masking step to separate function, think this is more or less it. Step 1 and Step 3 are cmd so if result.returncode != 0 we will get error message where the error occured

isce2_topsapp/iono_proc.py
mgovorcin: solved

isce2_topsapp/iono_proc.py
mgovorcin: added link to the original isce2 script

isce2_topsapp/iono_proc.py
mgovorcin: I made it like this to be able to update description of progress bar, probably there is nicer solution than this one

cmarshak:
Hey @mgovorcin - this looks good. Let's get this done today. Do the following:

1. Provide a sample CLI run
2. Put the output GUNW from the CLI run in a public bucket and link here

We can have @dbekaert  review quickly. If there are other pieces that you want to update, you can do so in a later PR.

Great work.

mgovorcin:
@cmarshak I tested it out twice at trappist, it runs without issues:

This is the command I used
```
isce2_topsapp --secondary-scenes S1B_IW_SLC__1SDV_20211108T014900_20211108T014928_029490_0384FB_3325 S1B_IW_SLC__1SDV_20211108T014926_20211108T014953_029490_0384FB_029B S1B_IW_SLC__1SDV_20211108T014950_20211108T015017_029490_0384FB_5476 S1B_IW_SLC__1SDV_20211108T015015_20211108T015042_029490_0384FB_1C30 --reference-scenes S1B_IW_SLC__1SDV_20211120T014900_20211120T014928_029665_038A5E_F809 S1B_IW_SLC__1SDV_20211120T014925_20211120T014952_029665_038A5E_FF40 S1B_IW_SLC__1SDV_20211120T014950_20211120T015017_029665_038A5E_5DE5 S1B_IW_SLC__1SDV_20211120T015015_20211120T015042_029665_038A5E_8246 --estimate-ionosphere-delay True
```
This enables using water mask in the ionospheric correction, however there are still some work that needs to be done for the refinement, which we can leave for another PR (as you mentioned)

cmarshak:
Looks great! Thanks, Marin.

image

cmarshak (approved):

PR #124, opened 2023-03-29
Solid Earth Tides: Pass coordinates from Geo Transforms to Pysolid, not lat/lon metadata arrays

sssangha:
Following our convo @cmarshak , I adjusted the logic of the SET code to handle and pass coordinates derived from the geotrans, as opposed to the latitude and longitude metadata arrays.

### DIFF
--- a/isce2_topsapp/solid_earth_tides.py
+++ b/isce2_topsapp/solid_earth_tides.py
@@ -66,9 +66,7 @@ def get_gunw_attrs(gunw_path: str) -> Tuple[float, np.array, np.array,
     group = 'science/grids/imagingGeometry'
     with xr.open_dataset(gunw_path, group=group, engine='rasterio') as ds:
         z_meta = ds.heightsMeta.data
-        lat_meta = ds.y.data
-        lon_meta = ds.x.data
-        gt  = ds.rio.transform()
+        gt = ds.rio.transform()
         # convert angles to rad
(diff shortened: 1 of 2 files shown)


isce2_topsapp/solid_earth_tides.py
```diff
@@ -66,18 +66,17 @@ def get_gunw_attrs(gunw_path: str) -> Tuple[float, np.array, np.array,
...
         solidtide_atr = {
-            'LENGTH': len(lat_meta),
-            'WIDTH': len(lon_meta),
-            'X_FIRST': lon_meta[0],
-            'Y_FIRST': lat_meta[0],
-            'X_STEP':  lon_meta[1] - lon_meta[0],
-            'Y_STEP': lat_meta[1] - lat_meta[0],
+            'LENGTH': az_angle.shape[1],
+            'WIDTH': az_angle.shape[2],
+            'X_FIRST': gt.c,
```
cmarshak, line 76: Thanks for doing this - I think it will be easier for someone to look up conventions in [`affine`](https://github.com/rasterio/affine), then how the string is designed.

cmarshak (changes requested):
Looks good; please change the final variable to float32 as that is what we generally use in the rest of the product. Also update version in changelog.

cmarshak (approved):

PR #122, opened 2023-03-27
Custom event

cmarshak:
Provides CLI options for:
  - output resolution: 30 meters or 90 meters (the latter is default and the standardized GUNW resolution)
  - dense offset layers: azimuth and range offset estimates as separate layers
  - unfiltered coherence layer

   If the output resolution is changed to 30 meters or any of the latter two layers are requested, then the initial token on product name is change from `S1-GUNW` to `S1-GUNW-CUSTOM`.

PR #128, opened 2023-04-05
Increase DEM buffer for Frames

cmarshak:
In our frame workflow, I have a geojson that estimates the GUNW footprint intersecting our latitude aligned frames with bursts. This is then what we use to determine how much DEM to request (this is the only place the footprints are used). The GUNW will always beyond the latitude aligned frame extent because ISCE2 selects all bursts that intersect the specified region of interest. So again there are two geometries that we are utilizing:

1.	The latitude aligned frame – determines what to feed into ISCE2’s region of interest (in the xml)
2.	The estimated extent – determines how much DEM to request for geocoding

As expected, there are slight discrepancies between what I can estimate using the ESA burst map and what ISCE2 actually does when selecting bursts. My guess is that this is due to the precise position of bursts on the ground during an acquisition which slightly differs from the ESA burst map and/or geometric manipulations within the footprint estimate (such as geometric simplifications of the latitude aligned frame for easier storage/io) that underscore unexpected side-effects that are not needed for further review at this time. 

That said, where this could be concerning (and this is the only place) is “are we requesting enough DEM for ISCE2 processing with our specified ROI?” and I want the answer to be an emphatic "no". Currently, I would say I think we could potentially be in a weird edge case. Specifically, the problem is that we could potentially have extra bursts included by ISCE2 and not enough DEM to geocode.

So, let's put this aside and increase the extent buffer to .4 which amounts to ~40 km (or 2 bursts) around the estimated extent and this will always ensure we have enough DEM (there will never be more than 1 burst discrepency from the estimated footprint). Note we further round to integer degrees so we are often requesting a lot, lot more. https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/dev/isce2_topsapp/localize_dem.py#L67-L70

Again, this change is to ensure that we ask for sufficient amount of DEM everywhere. This is to be extraordinarily cautious and to ensure we are not having weird edge cases.

### DIFF
--- a/isce2_topsapp/localize_dem.py
+++ b/isce2_topsapp/localize_dem.py
@@ -47,7 +47,7 @@ def download_dem_for_isce2(extent: list,
     Parameters
     ----------
     extent : list
-        [xmin, ymin, xmax, ymin] for epsg:4326 (i.e. (x, y) = (lon, lat))
+        [xmin, ymin, xmax, ymax] for epsg:4326 (i.e. (x, y) = (lon, lat))
     dem_name : str, optional
         See names in `dem_stitcher`
     full_res_dem_dir : Path, optional


isce2_topsapp/localize_dem.py
```diff
@@ -42,7 +42,7 @@ def download_dem_for_isce2(extent: list,
                            dem_name: str = 'glo_30',
                            full_res_dem_dir: Path = None,
                            low_res_dem_dir: Path = None,
-                           buffer: float = .1) -> dict:
+                           buffer: float = .4) -> dict:
     """
     Parameters
     ----------
```
jtherrmann, line 48: Will want to update the docstring description for `buffer`.

jtherrmann: Also, this may be a good opportunity to note that the second `ymin` should be changed to `ymax` in the docstring description for `extent`.

jtherrmann: Thanks! I was looking at this code because ASF plans to utilize it as part of our [HyP3 ISCE2 plugin](https://github.com/ASFHyP3/hyp3-isce2).

cmarshak: Please take a look at this too: https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/pull/122

This has some of the features that may be particularly relevant for on-demand processing including:

1. unfiltered coherence
2. 30 meter resolution

I think we are also discussing putting in Reference and Secondary Backscatter as additional layers. Hope this helps. Glad this is (potentially) being integrated as on-demand processing.  If the stars align maybe we can share some basic hyp3 tests or piggy back off your dependabot. The ISCE2 repo dependencies often give me a 🤯.

jtherrmann: Thanks, we'll take a look. @forrestfwilliams and I were just discussing how we might reduce duplication between the two projects so I will run all of this by him as well.

forrestfwilliams: Yes, definitely hope we can combine efforts!

mgovorcin (approved):

PR #130, opened 2023-04-18
Iono processing update

mgovorcin:
- pythonized ionosphere computation workflow, e.g. using topsapp runIon module instead of running xml template with command prompt
 - new changes to the topsapp default iono workflow:
 - 1. mask connected component 0 after unwrapping (removes noisy pixels along the coastline)
 - 2. apply simple "phase bridging" between connected components lables
 - 3. modified gaussian adaptive filtering, mask out zeros (nan values) before calculating statistics
 - 4. added option to estimate burst phase jumps, at the moment is set to False as is missing triggering function, this is the fix for [isce2 #620](https://github.com/isce-framework/isce2/issues/620)
 -5. include changes to ionosphere computation workflow swath by swath workflow (gets called out for cross-S1-A/B interferogram)  

GUNW example: https://gunw-development-testing.s3.us-west-2.amazonaws.com/iono_wmask_testing/S1-GUNW-A-R-064-tops-20211120_20211108-014951-00119W_00030N-PP-aed8-v2_0_6/S1-GUNW-A-R-064-tops-20211120_20211108-014951-00119W_00030N-PP-aed8-v2_0_6.nc


cmarshak (dismissed):
I leave it to you if you want to integrate my suggestions. I realize that the code itself is directly from ISCE2 scripts. Anyways, this looks really good. My comments are aesthetic. It takes a lot to understand these internals and modify them. Great work.

cmarshak:
Hey @mgovorcin - can you add some basic control flow to ensure ESD is -1 when ionosphere is turned on?

@dbekaert says we don't need to ever do both based on this comment: https://github.com/isce-framework/isce2/discussions/705#discussioncomment-5703707

cmarshak:
Also - FYI - I won't review until you ask - but currently the GUNW you shared above is ~200 MBs. Not sure why it's so large. I would definitely double check with David about this.

mgovorcin:
> 
@cmarshak I found out that the size ~200mb is due to stitched GUNW product (topsApp used 4 ref and second SLC scenes). When checking the GUNW produced by using fixed frame id, the size is around 52mb .

Also, I added the code to swithc off ESD when using iono in main.py

dbekaert:
Would expect it to be in range of like 60-80MB or so for a frame.

cmarshak:
Thanks for all your hard work on this - let's test this via hyp3 test jobs in the coming days.

I realized some of my comments like "change the name of `mergeBurst2`" is moot because you are importing this.

cmarshak (approved):
I think you added some more code directly from ISCE2 scripts. Hope there are enough pointers. Thanks again.

PR #131, opened 2023-04-26
Solid Earth Tide using per pixel Azimuth Time derived from ISCE2 orb.geo2rdr

cmarshak:
Solid Earth Tide computation using the `orb.geo2rdr` which outputs a tuple `azimuth_time` and `range` as noted [here](https://github.com/isce-framework/isce2/blob/c2d1a218020d81071574cd70033b7d673e2f5092/components/isceobj/Orbit/Orbit.py#L1000) shared by @mgovorcin.

Steps are:

1. Get azimuth time array with coordinate frame determined by `science/grids/imagingGeometry
2. Compute Solid Earth tide for the azimuth time rounded to nearest second above and below precise azimuth time and compute a weighted average.
3. Package into an xarray georeferenced dataset

The [notebook](https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/SET-azimuth-time/tests/solid_earth_tides.ipynb) test is for interacting with the output and the pytest test file ensures georferencing is done correctly so can be ready by `gdal` or `rasterio`.


tests/test_set.py
```diff
@@ -0,0 +1,58 @@
...
+    # Azimuth time array
+    X = get_azimuth_time_array(product_dir=isce_data_dir_for_set / 'reference',
+                               height_mesh_arr=hgt_mesh,
+                               latitude_mesh_arr=lat_mesh,
+                               longitude_mesh_arr=lon_mesh)
+    # Total seconds from minimum time
+    Y = (X - X.min()) / np.timedelta64(1, 's')
+    # Vertical standard deviation
+    Y_std = Y.std(axis=0)
+    assert_almost_equal(Y_std, 0, decimal=3)
```
cmarshak, line 58: @dbekaert - I was able to show that in a vertical column of the azimuth time array that we comute has a standard deviation that never exceeds `1e-3` seconds.

However, if the heights in the `imagingGeometry` group have a max of 9,000 m and min of 1,500 m -- the speed of light takes 3.5e-5 seconds to travel between the highest and lowest level in the cube approximately (or 2 orders of magnitude smaller than the maximum standard deviation in this test).

dbekaert: That sounds about right and negligible. We should just make a note of it in the code that this is the assumption we do. If we dont do it thats also ok. There overhead should be small.

cmarshak:
Here are some sample products with overlapping spatial/temporal area: [GUNW 0](https://gunw-development-testing.s3.us-west-2.amazonaws.com/solid_earth_tide_may_2023/S1-GUNW-A-R-064-tops-20211120_20211108-014926-00119W_00032N-PP-c832-v2_0_6.nc) and [GUNW 1](https://gunw-development-testing.s3.us-west-2.amazonaws.com/solid_earth_tide_may_2023/S1-GUNW-A-R-064-tops-20211120_20211108-014951-00119W_00033N-PP-0643-v2_0_6.nc).

To update a product with the solid earth tide yourself, here is the sample code:

```
from pathlib import Path
import xarray as xr
from hyp3lib import get_orb
from isce2_topsapp.solid_earth_tides import update_gunw_with_solid_earth_tide

nc_path = Path('<GUNW_PATH>')
group = f'science/radarMetaData/inputSLC/reference'
with xr.open_dataset(nc_path, group=group) as ds:
    slc_id_ref = ds['L1InputGranules'].values[0]
group = f'science/radarMetaData/inputSLC/secondary'
with xr.open_dataset(nc_path, group=group) as ds:
    slc_id_sec = ds['L1InputGranules'].values[0]

orb_file_ref, _ = get_orb.downloadSentinelOrbitFile(slc_id_ref)
orb_file_sec, _ = get_orb.downloadSentinelOrbitFile(slc_id_sec)

update_gunw_with_solid_earth_tide(nc_path, 'reference', [orb_file_ref])
update_gunw_with_solid_earth_tide(nc_path, 'secondary', [orb_file_sec])
```

For the API, I have a list of orbit files, but I don't think that is necessary. May want to have single path or string.

@sssangha - take a look at it when you get back and then let Marin give a final review after you make any necessary modifications.

cmarshak:
There is an issue that was discovered by @sssangha with this PR.

Unsure how this is normally dealt with.

It appears that `orb.geo2rdr` requires more steps to converge to the azimuth time and rng distance when using the *full* orbit file. In  ISCE3 there are options to increase the number of iterations of the numerical method to approximate this and in Raider there is a clipping of the orbit around is reference acquisition time. Because ISCE2 does not have options to increase the number of iterations we will pursue the latter.

cmarshak:
@sssangha - this is ready for review - the same code snippet above still works.

I also added a test that I will link to below that resolves the issue you caught regarding geo2rdr.

tests/test_set.py
```diff
@@ -0,0 +1,79 @@
...
+
+    # Ensure that azimuth times are all within specified padding used in computation
+    # Padding used is 600 seconds - we check for 70 - note we are using the SLC id start time
+    # From secondary and there are TWO slcs so this makes sense with an SLC requiring being acquired
+    # over about 20-30 seconds; also the meshgrid for this layer is slightly larger than the GUNW as well
+    # indicating why the 10 second buffer is used.
+    x = pd.to_datetime(X.ravel())
+    differences_in_sec = (x - slc_start_time).total_seconds()
+    result = np.all(differences_in_sec < 70)
+    assert result
```
cmarshak, line 79: @sssangha - ensures that the azimuth timing array never deviates from the first SLC start time by more than 70 seconds. For the GUNW under inspection, there are two secondary images (which are used above) and that's why we use the 70. Let me know if this looks good.

sssangha:
@cmarshak I fixed the shift of lat array (previously the factor was added in this [line](https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/SET-azimuth-time/isce2_topsapp/solid_earth_tides.py#L177)) vis-a-vis the mintpy convention [here](https://github.com/insarlab/MintPy/blob/main/src/mintpy/prep_aria.py#L183).

cmarshak:
@sssangha - I am reverting this.

Note in the mintpy is using the *affine transform* where we are using positive resolutions. The `+` was correct. Mintpy is using the signed resolutions as noted earlier in its code: https://github.com/insarlab/MintPy/blob/main/src/mintpy/prep_aria.py#L181

cmarshak:
@sssangha - please approve once you are ready. After that, David can also review with you.

Thanks!

tests/test_set.py
```diff
@@ -0,0 +1,170 @@
...
+    path_to_set = (f'netcdf:{tmp_gunw}:/science/grids/corrections/external/'
+                   f'tides/solidEarth/{acq_type}/solidEarthTide')
+    with rasterio.open(path_to_set) as ds:
+        X = ds.read()
+
+    X_set_plugin_mm = X * 0.055465761572122574 / np.pi / 4 * 1_000
+    X_set_pysolid_mm = get_pysolid_set(tmp_gunw, acq_type=acq_type)
+
+    set_abs_diff = np.abs(X_set_pysolid_mm - X_set_plugin_mm)
+    assert np.max(set_abs_diff) < 1
```
cmarshak, line 170: @dbekaert - FYI - this is a test that Sim and I worked on that ensures that the pysolid grid (used in mintpy) and the time varying interpolation used here do not vary by more than 1 mm across the scene.

tests/test_set.py
```diff
@@ -0,0 +1,179 @@
...
+        path_to_set = (f'netcdf:{tmp_gunw}:/science/grids/corrections/external/'
+                       f'tides/solidEarth/{acq_type}/solidEarthTide')
+        with rasterio.open(path_to_set) as ds:
+            X = ds.read()
+
+        X_set_plugin_mm = X * 0.055465761572122574 / np.pi / 4 * 1_000
+        X_set_pysolid_mm = get_pysolid_set(tmp_gunw, acq_type=acq_type)
+
+        set_abs_diff = np.abs(X_set_pysolid_mm - X_set_plugin_mm)
+        assert np.max(set_abs_diff) < 1
```
cmarshak, line 179: @dbekaert - (had to remove previous comment because I updated the test) - this includes a verification that pysolid using fixed time does not deviate by more than 1 mm than the existing plugins method. This is something Sim and I worked on. This test will verify this for two datasets within the test suite and for both the reference and secondary dates in each dataset.

isce2_topsapp/solid_earth_tides.py
```diff
@@ -6,96 +6,382 @@
...
+                                   position=position,
+                                   velocity=velocity)
+        state_vector_list.append(state_vector)
+
+    return state_vector_list
+
+
+def get_orbit_obj_from_orbit_xmls(orbit_xmls: list[Path],
+                                  slc_start_time: datetime.datetime,
+                                  pad_in_seconds: int = 600) -> Orbit:
```
mgovorcin, line 92: wonder how is the padding set in the isce topsAp, is it also 600 seconds?

cmarshak: No idea...

Sim was able to track this down: https://github.com/dbekaert/RAiDER/blob/dev/tools/RAiDER/losreader.py#L188-L200

My rough understanding is that if there is not a padding then there are two solutions occurring at antipodal points on the globe.

isce2_topsapp/solid_earth_tides.py
```diff
@@ -6,96 +6,382 @@
...
-        gt = ds.rio.transform()
-        # convert angles to rad
+    with xr.open_dataset(gunw_path, group=group, mode='r') as ds:
+        # latitude resolution will match the rasterio transformed; i.e. it will be negative
+        lon_res, lat_res = ds.rio.resolution()
+        # We make the resolution positive
+        lat_res = -lat_res
+
+        height_coord_arr = ds.heightsMeta.data
+        # half pixel shift to ensure AREA (i.e. upper left corner) convention assumed by pysolid
```
mgovorcin, line 177: it think calculation of SET should be done for the pixel centroid rather than corner, not sure how pysolid calcs this as it accepts AREA style indeed. Is there anywhere in the code where pysolid does half pixel shift in both directions

mgovorcin (approved):
Looks great!, added two small comments for the discussion.

isce2_topsapp/solid_earth_tides.py
cmarshak: Not totally clear on what you are asking here.  

The python wrappers for the fortran code indeed want the `AREA` formatted metadata (as you mention) which would align with mintpy's attribute `dict`[https://github.com/insarlab/PySolid/blob/main/src/pysolid/grid.py#L30-L33]. I am assuming the georeference/gdal metadata of the outputs would align with those of the inputs (maybe this is where your question is?). This pysolid notebook is pretty clear about the inputs (not about the outputs): https://github.com/insarlab/PySolid/blob/main/docs/plot_grid_SET.ipynb

Also the other GUNW layers specifically the geometry rasters follow the convention:

1. Gdal metadata is `AREA`
2. The coordinates are `PIXEL`
3. The raster is consistent with 1.

This is what I have seen.

PR #132, opened 2023-05-15
Add option to control OpenMP threads

### LINKED ISSUE #106: Add option to set optimum number of threads to `main()` entrypoint
ISCE2 uses OpenMP threads for parallelization which can be controlled via an `OMP_NUM_THREADS` environment variable. 

Setting this variable is important to:
* prevent CPU contention when running multiple processing "jobs" on the same machine
* reduce runtime, especially in the topo step, by better utilizing the available resources

When manually running the DockerizedTopsApp container, `OMP_NUM_THREADS` can be provided as an environment variable. For HyP3 processing jobs, we'd prefer to expose this as an option for better transparency. We've done this already for our other plugins -- ASFHyP3/hyp3-gamma#428 can provide a good template for doing it here. 

Notably, we we set the environment variable in `main()` before we load any of the processing entrypoints to ensure it gets picked up appropriately.

jhkennedy:
See description in issue :point_down: 

fixes #106

### DIFF
--- a/isce2_topsapp/__main__.py
+++ b/isce2_topsapp/__main__.py
@@ -335,9 +335,9 @@ def main():
     parser.add_argument(
         "++omp-num-threads",
         type=int,
-        help=("The number of OpenMP threads to use for parallel processing; " 
-              "when running locally, this routine will utilize all resources; "
-              "suggest to use 8 to 16 so other processes can running locally.")
+        help=("The number of OpenMP threads to use for parallel processing in ISCE2 routines; "
+              "when running locally, this topsapp will utilize all resources, which is not recommended; "
+              "suggest to set this option to 8 - 16 so other processes on server/workstation can running.")
     )
 
     args, unknowns = parser.parse_known_args()
--- a/isce2_topsapp/__main__.py
+++ b/isce2_topsapp/__main__.py
@@ -335,7 +335,9 @@ def main():
     parser.add_argument(
         "++omp-num-threads",
         type=int,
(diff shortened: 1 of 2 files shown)


cmarshak (dismissed):

cmarshak (dismissed):

isce2_topsapp/__main__.py
```diff
@@ -332,8 +332,17 @@ def main():
         default="gunw_slc",
         help="Select the HyP3 entrypoint to use",
     )
+    parser.add_argument(
+        "++omp-num-threads",
+        type=int,
+        help="The number of OpenMP threads to use for parallel processing; when running locally, this routine will utilize all resources; suggest to use 8 to 16 so other processes can running locally."
+    )
+
```
cmarshak, line 340:
```suggestion
    parser.add_argument(
        "++omp-num-threads",
        type=int,
        help=("The number of OpenMP threads to use for parallel processing; " 
                   "when running locally, this routine will utilize all resources; "
                   "suggest to use 8 to 16 so other processes can running locally.")
    )

```

cmarshak:
@jhkennedy - before we merge - wanted your input on this: https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/resource-competition/isce2_topsapp/iono_proc.py#LL36C1-L39C21

This changes the available threads to 4 if it does not exist. This could yield unexpected results. We might want to delete this and/or make the default value 4.

cmarshak (approved):

PR #133, opened 2023-05-22
Fix localize SLC tests

cmarshak:
Enforcing that no warning was thrown was causing an error when a new `RuntimeWarning` was raised not related to our particular function call.

Will be good to revisit.

Also, using Natural Earth url as this dataset will not be distributed in the future as indicated here: https://github.com/geopandas/geopandas/issues/2751


cmarshak:
@mgovorcin - I am turning on auto merge.

Please review. Once it's merged into dev. We can intergrate this back into your branch so the test suite passes.

mgovorcin (approved):

PR #138, opened 2023-06-06
Iono computation - addressing #135

mgovorcin:
Changes:
1. Skip iono_computation if lower/upper band unwrapped interferogram array is empty, i.e. all zeros. Return iono as zeros array
2. Skip downloading and using water mask if the AOI request is out of SWBD data coverage. TODO: introduce new water masks as mentioned in #135 : EXAMPLES: https://worldcover2021.esa.int/viewer or https://land.copernicus.eu/global/products/wb
3. fix few typos and add notes in the __main__ regarding the esd and calculation of burst jumps

### DIFF
--- a/isce2_topsapp/__main__.py
+++ b/isce2_topsapp/__main__.py
@@ -58,7 +58,7 @@ def localize_data(
         # For ionospheric correction computation
         if water_mask_flag:
             out_water_mask = download_water_mask(
-                out_slc["extent"],  water_name='SWDB')
+                out_slc["extent"],  water_name='SWBD')
 
         out_aux_cal = download_aux_cal()
 
--- a/isce2_topsapp/__main__.py
+++ b/isce2_topsapp/__main__.py
@@ -178,6 +178,10 @@ def gunw_slc():
               indent=2, cls=MetadataEncoder)
 
     # Turn-off ESD when using ionospheric computation
+    # NOTE: note sure if this needs to be off
+    #       esd is calculate with iono only when
+    #       considerBurstProperties is on which off 
+    #       by default
     if args.estimate_ionosphere_delay:
         args.esd_coherence_threshold = -1
 
@@ -201,7 +205,8 @@ def gunw_slc():
     #   this option (it adds 10min to iono for frame processing)
     #   example: look at filt_toposphase.unw or .flat
     #   and analyze if there are any burst jumps,
-    #   if yes, set this option True
+    #   if yes, set this option True, or we could run it always
+    #   and store both iono-long wavelength and burst jumps(az_shifts)
     if args.estimate_ionosphere_delay:
         iono_processing(
             mask_filename=loc_data["water_mask"],
--- a/isce2_topsapp/__main__.py
+++ b/isce2_topsapp/__main__.py
@@ -198,7 +198,7 @@ def gunw_slc():
     # Run ionospheric correction
     # MG: correct burst jumps, e.g. needed for Arabian
     #   processing. TODO: We need a trigger function for
-    #   this option (it adds almost double time to iono)
+    #   this option (it adds 10min to iono for frame processing)
     #   example: look at filt_toposphase.unw or .flat
     #   and analyze if there are any burst jumps,
     #   if yes, set this option True
--- a/isce2_topsapp/__main__.py
+++ b/isce2_topsapp/__main__.py
@@ -179,8 +179,8 @@ def gunw_slc():
 
     # Turn-off ESD when using ionospheric computation
     # NOTE: note sure if this needs to be off
-    #       esd is calculate with iono only when
-    #       considerBurstProperties is on which off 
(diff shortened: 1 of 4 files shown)


isce2_topsapp/__main__.py
```diff
@@ -178,6 +178,10 @@ def gunw_slc():
               indent=2, cls=MetadataEncoder)
 
     # Turn-off ESD when using ionospheric computation
+    # NOTE: note sure if this needs to be off
```
cmarshak, line 181:
```suggestion
    # NOTE: not sure if this needs to be off.
```
Fix typo

isce2_topsapp/localize_mask.py
```diff
@@ -19,17 +19,27 @@ def download_water_mask(
...
-            extent_buffered[1],
-            extent_buffered[3],
-            extent_buffered[0],
-            extent_buffered[2],
-        )
+        # Water mask dataset extent
+        # Latitude S55 - N60
+
+        lats = [extent_buffered[1], extent_buffered[3]]
+        if (np.abs(lats) < 59.9).all():
```
cmarshak, line 28: Just to clarify, the lats > -56 (not sure if the remaining 3 degrees in Southern hemisphere is just water)

```suggestion
        if ((lats < 59.9).all()) and ((lats > -55.9).all()):
```

isce2_topsapp/localize_mask.py
```diff
@@ -19,17 +19,27 @@ def download_water_mask(
...
+                extent_buffered[1],
+                extent_buffered[3],
+                extent_buffered[0],
+                extent_buffered[2],
+            )
+            mask_filename = str(output_dir / mask_filename)
+        else:
+            print('Request out of SWBD coverage ',
+                  'Skip downloading water mask!!')
+            mask_filename = ''
```
cmarshak, line 39: Can you please clarify why you don't raise an error? If there is no water mask available, will ionosphere process run to completion yielding an all zero layer? Or will something else occur?

I think we may want to raise an error in this case because ionosphere should not be computed right? However, I do like the idea that we can create generic products beyond a water mask coverage area. What are your thoughts.

Generally, operator (person submitting jobs on hyp3) will avoid these higher/lower latitudes.

isce2_topsapp/iono_proc.py
```diff
@@ -1132,3 +1120,88 @@ def ionSwathBySwath(self: Type[topsApp.TopsInSAR],
...
+              ' skip iono computation!')
+        ionos = np.zeros_like(lowerUnw)
+    else:
+        # compute ionosphere
+        fl = runIon.SPEED_OF_LIGHT / ionParam.radarWavelengthLower
+        fu = runIon.SPEED_OF_LIGHT / ionParam.radarWavelengthUpper
+        # polynomial method for removing relative phase unwrapping errors
+        adjFlag = 1
+        ionos = runIon.computeIonosphere(lowerUnw, upperUnw, cor, fl, fu,
+                                         adjFlag, corThresholdAdj, 0)
```
cmarshak, line 1185: I guess this is the important piece that will fix the swath issues.

To be clear, if a swath is all zeros but a swath nearby is not will the ionosphere look ok merged? We probably want to show some examples of Chile here to David.

cmarshak (changes requested):
Looks great. Requested some small changes/clarification. Otherwise, let's merge soon so we can test.

isce2_topsapp/localize_mask.py
mgovorcin: SWBD coverage goes up to S54 deg, below data is only water so do not think we are going to do any processing there

it was just easier to write it like this using abs.

isce2_topsapp/localize_mask.py
mgovorcin: There is no error as ionosphere computation will go on without using water mask - it will return to default processing with minor modifications. 
1. iono will result in all zero only if there is no land/or data to compute iono
2. we do not need water mask to do iono computation, we use it to improve iono near coastline where mask is available. To be more specific, we are masking water out before unwrapping lower and upper band
3. I think we can still run high latitudes with respect to water mask which is optional, but we do need dem. I would suggest to make sure that DEM coverage is documented is Warning is raised if it is out of its scope.

isce2_topsapp/iono_proc.py
mgovorcin: it looks good based on the local run, anyways we do not care about swath over water as that data is not to be trusted anyways. In this case we are going to get additional WARNING that code could not perform swath adjustment. But we can ignore it over water

cmarshak (approved):

dbekaert:
@mgovorcin Should have meta data to the iono correction  in gunw to track how it was produced

Water mask source : e.g none, gssh, …
Method: e.g none, framebased, subswath

Then we can always track it in the product. Please add in next pr

PR #141, opened 2023-06-26
adding iono processing attributes to GUNW

mgovorcin:
Adding ionosphere processing additional attribute to GUNW packging of ionosphere correction layer.

- Added attributes to ionosphere GUNW layer metadata:
1. ionosphere **processing_steps** - list:
a) iono-long wavelength: ['subband', 'rawion', 'filt_gaussian', 'geocode'],  
b) iono-long wavelength+az_shift per bursts: ['subband', 'rawion', 'filt_gaussian''ionosphere_shift', 'ion2grd', 'esd', 'geocode']
2. **water_mask** filename   str -. e.g. swbdLat_N47_N50_Lon_W072_W067.wbd (SRTM SWBD water mask)  or '' for None
3. **mask_connected_component_zero** flag - str(True/False): Mask out unwrapping connected component 0 after unwrapping lower and upper band
4. **do_phase_bridging** flag  - str(True/False): apply simple phase bridging between unwrapped components to minimize unwrapping error, i.e., 2pi jumps
5. **estimate_burst_jumps** flag - str(True/False): estimate burst jumps, aka azimuth shift per burst (scalloping effect - miscorregistration) caused by high ionosphere delay. NOTE depends on the quality of the long-wavelength ionosphere delay estimate from split-spectrum.
6. **burst_jumps_description** - str - description of this processing mode, as mentioned above
7. **swath_mode flag** - str(True/False): estimate ionosphere correction per swath, this mode get turn off for cross S1A/B interferogram pair, or when the starting range is different between reference and secondary image
8. **swath_ramp_removal** - str(True/False): remove ramps between swaths, adjustment in swath mode. By default is turn off
9.  **swath_mode_description** -- description of this processing mode, as mentioned above
10. **multilook_az_rg1** - list - multillooking ration in azimuth and range direction in the first round for iono compution
11. **multilook_az_rg2** - list - multillooking ration in azimuth and range direction in the second round for iono compution
12. **iono_height** - str - ionoshere height for the thin ionosphere shell in km, needed for projection from ground2iono which is turned off by default

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Added
 * localize_data within __main__.py added option to use/not use water mask for ionosphere processing
 * Added option to estimate burst phase jumps in ionosphere computation
+* Added packing of additional attributes for ionosphere computation into GUNW
 
 ## [0.2.3]
 
--- a/isce2_topsapp/__main__.py
+++ b/isce2_topsapp/__main__.py
@@ -208,7 +208,7 @@ def gunw_slc():
     #   if yes, set this option True, or we could run it always
     #   and store both iono-long wavelength and burst jumps(az_shifts)
     if args.estimate_ionosphere_delay:
-        iono_processing(
+        iono_attributes = iono_processing(
             mask_filename=loc_data["water_mask"],
             correct_burst_jumps=False,
         )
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Added
 * localize_data within __main__.py added option to use/not use water mask for ionosphere processing
 * Added option to estimate burst phase jumps in ionosphere computation
+* Added packing of additional attributes for ionosphere computation into GUNW
```
cmarshak, line 23: Can you please clarify what these attributes are here and in the PR?

cmarshak (commented):
I believe this adding additional attributes to the metadata layer. Please clarify. Might be nice to put in the PR an example of what the attributes actually say (as a dict) so there is a reference to this. It will be helpful to know particularly since there appears to be some small modifications.

Thanks for doing this.

CHANGELOG.md
mgovorcin: added to PR and changelog. Let me know if this is sufficient

cmarshak (approved):

PR #151, opened 2023-08-31
ensure creation_timestamp includes microseconds

asjohnston-asf:
ASF's ingest workflow requires datetimes in the delivery metadata file to be formatted `2023-01-01T00:00:00.000000Z` per the schema at https://github.com/asfadmin/grfn-ingest/blob/test/verify/src/metadata_schema.json

[datetime.datetime.isoformat()](https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat) does not include microseconds if the microsecond value is exactly zero:
```
>>> datetime.datetime(2023, 1, 1, 0, 0, 0).isoformat()
'2023-01-01T00:00:00'
>>> datetime.datetime(2023, 1, 1, 0, 0, 0, 1).isoformat()
'2023-01-01T00:00:00.000001'
```

This PR eliminates the one-in-a-million case where `creation_timestamp` did not include microseconds, causing ingest to ASF to fail. We saw one instance of this in the recent processing campaign.

### DIFF
--- a/isce2_topsapp/delivery_prep.py
+++ b/isce2_topsapp/delivery_prep.py
@@ -117,7 +117,6 @@ def gen_browse_imagery(nc_path: Path,
 def format_metadata(nc_path: Path,
                     all_metadata: dict) -> dict:
 
-    now = datetime.datetime.now()
     label = nc_path.name[:-3]  # removes suffix .nc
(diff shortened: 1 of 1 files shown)


cmarshak:
Thanks for the explanation and quick resolution.

cmarshak (approved):

PR #197, opened 2025-02-21
v0.4.0

ehavazli:
This is the orbit fix pulling data from ASF's S3 bucket. We will make sure the plugin works via test deployment and then resume processing.


cmarshak:
@ehavazli - this is ready to test using the `test` endpoint. Let's check in briefly after that.

I am very confident it will work after Andrew updated our test suite and did a single run. 

I am going to merge this PR after it passes the test to ensure no weird environment items changes. If you find issues, we can do another release.

(Note how I increased the label to `major`).

Thanks, @asjohnston-asf

cmarshak (approved):

PR #198, opened 2025-02-21
Adopt s1-orbits and remove esa credentials/hyp3lib 

asjohnston-asf:
Here's a more thorough orbit cleanup to consider. Using the s1-orbits package means hyp3lib is not longer needed as a dependency, and the esa username/password parameters are no longer required as inputs.

I don't fully appreciate the `dry_run` parameter; is it important that no orbit files be downloaded to disk when doing a dry run?

The test notebooks include code/instructions for setting up esa credentials and downloading orbits that still deserve to be updated. I'm opening the PR because I've been noodling on it today, but I'm not committed to doing the remaining work.


cmarshak:
Andrew - probably don't have much time to devote to updating the tests... and they are in notebooks. If you or even Emre can update these (they are probably trivial) that would be great.

I am going on leave in 2 weeks and trying to get everything done before then.

PR #199, opened 2025-02-21
Thorough update of `s1_orbits`

cmarshak:
@asjohnston-asf - thanks.

We definitely don't need to keep `dry-run` - we were doing the bare minimum to update the plugin. There may be some tests that are required - can't tell.

Let's see if the tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
-## [0.3.9]
-* Use s1_orbits for downloading orbit data
+
+## [0.4.0]
+
+### Changed
+* Use `s1_orbits` rather than `hyp3lib` to download Sentinel-1 orbits files in `localize_orbits.download_orbits`
+
+### Removed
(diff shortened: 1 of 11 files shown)


cmarshak:
@asjohnston-asf - there are some linting issues and whatnot. If it passes, we can merge to `dev`.

@ehavazli - wait until this is merged to dev to make the small test run.

Then we can release - we would have to change from `patch` to `minor` bump. This amounts to changing the label.

asjohnston-asf:
Tests are passing, and I was able to generate `S1-GUNW-D-R-036-tops-20250131_20241226-041630-00025E_00035N-PP-99eb-v3_0_1` using the `ghcr.io/access-cloud-based-insar/dockerizedtopsapp:0.3.9.dev12_g8750e0b` container built from this PR.

`isce2_topsapp --reference-scenes S1A_IW_SLC__1SDV_20250131T041604_20250131T041631_057683_071C03_846A S1A_IW_SLC__1SDV_20250131T041629_20250131T041656_057683_071C03_79E2 --secondary-scenes S1A_IW_SLC__1SDV_20241226T041607_20241226T041634_057158_07073F_4DAB S1A_IW_SLC__1SDV_20241226T041632_20241226T041659_057158_07073F_315A --frame-id 5495`

I haven't tried the gunw burst workflow; I'm not familiar with it.

cmarshak:
Thanks for doing this, @asjohnston-asf - super appreciate it! Not just for updating the plugin, but for making the library and hopefully removing this headache.

cmarshak:
I am merging.

PR #205, opened 2025-05-12
Job configuration with dedup for topsapp based on frame-id, ref and sec date

mgovorcin:
**Added new entrypoint to generate topsapp configuration dict** 
`topsapp_config --frame-id 9848 --reference-date 20150502 --secondary-date 20250520`

Wrong date input:
`topsapp_config --frame-id 9848 --reference-date 2015-05-02 --secondary-date 2015-05-20`

GUNW exists:
`topsapp_config --frame-id 9848 --reference-date 2015-05-02 --secondary-date 2015-05-26`

Output:
```
{'reference': ['S1A_IW_SLC__1SDV_20250509T015010_20250509T015040_059111_07557C_C6AD',
  'S1A_IW_SLC__1SDV_20250509T015038_20250509T015105_059111_07557C_191F'],
 'secondary': ['S1C_IW_SLC__1SDV_20250503T014857_20250503T014924_002160_004961_E4B4',
  'S1C_IW_SLC__1SDV_20250503T014922_20250503T014949_002160_004961_D210'],
 'reference_date': '2025-05-09',
 'secondary_date': '2025-05-02',
 'frame_id': 9848}
```

Some code feature:
 - code infers reference date as latter date from specified dates (e.g. switch second with ref, if secondary is latter date)
 -  code drops Sentinel1 SLC that fail to cover 95% of the frame area
 -  code raises error if wrong reference and/or secondary date have be specified, and points to nearest date around
 - code raises ValuError if S1_ARIA_GUNW already exists in the DAAC

ToAdd:
 - logging: Dockerized Tops does not have log module so did not go to add one, but some of this raise Value could be replace with logging message 
 - output format (TBD) for now dict is send to STOUT, will wait to on the feedback how this can be implement for ondemand

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -99,6 +99,7 @@ conf/settings.conf
 tests/*.ipynb
 tests/out/*.ipynb
 !tests/localize-data.ipynb
+!tests/config-job.ipynb
 /isce2_topsapp.egg-info/
 !tests/prepare-for-delivery.ipynb
 !tests/solid_earth_tides.ipynb
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,43 @@
+ci:
+  autofix_prs: false
+
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: "v5.0.0"
+    hooks:
+      # https://github.com/pre-commit/pre-commit-hooks/issues/718
+      # - id: check-added-large-files  # Fails with git v1.8.3
+      - id: check-case-conflict
+      - id: check-docstring-first
+      - id: check-merge-conflict
+      - id: check-yaml
+        args: [--allow-multiple-documents]
+      - id: debug-statements
+      - id: end-of-file-fixer
+      - id: file-contents-sorter
+        files: (requirements.txt)$
+      - id: mixed-line-ending
+      - id: trailing-whitespace
+
+
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.11.8
+    hooks:
+      - id: ruff
+        args: [--fix, --exit-non-zero-on-fix, --show-fixes]
+
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: "v1.15.0"
+    hooks:
+      - id: mypy
+        additional_dependencies:
+        - types-setuptools
+        - types-requests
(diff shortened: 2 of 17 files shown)


mgovorcin:
@asjohnston-asf  please do check and give your feedback, especially I am interested in how you envision this to run ondemand and what should be the output

asjohnston-asf:
First, if you're going to do a code formatting update, please do that in a separate PR from the feature changes. I don't have the mental bandwidth to look through 2,000 changed lines.

sssangha:
> `topsapp_config --frame-id 9848 --reference-date 20150502 --secondary-date 20250520`

This doesn't work, but based on the dictionary based below I assume the specified `--reference-date` is `20250502` `--secondary-date` is `20250509`.

I can confirm this works as intended:
`topsapp_config --frame-id 9848 --reference-date 20250502 --secondary-date 20250509`

asjohnston-asf:
I think this PR misses the mark somewhat. On-demand requests currently invoke DockerizedTopsApp using parameters like:
`isce2_topsapp --bucket myBucket --bucket-prefix myPrefix --frame-id 24299 --reference-scenes S1A_IW_SLC__1SSV_20141227T100440_20141227T100507_003903_004AFB_4A35 S1A_IW_SLC__1SSV_20141227T100505_20141227T100532_003903_004AFB_8FEF --secondary-scenes S1A_IW_SLC__1SSV_20141203T100441_20141203T100508_003553_004301_CF13 S1A_IW_SLC__1SSV_20141203T100505_20141203T100533_003553_004301_71CB`

Which results in `S1-GUNW-D-R-156-tops-20141227_20141203-100506-00075W_00038S-PP-5c1c-v3_0_1` being uploaded to the specified S3 bucket.

The first feature we're trying to deliver is to provide an alternate (or enhanced) entrypoint that allows the user to specify dates instead of scenes, e.g.:

`isce2_topsapp --bucket myBucket --bucket-prefix myPrefix --frame-id 24299 --date1 2014-12-27 --date2 2014-12-03`

And have the same interferogram still generated and uploaded to S3.


We should defer implementing de-duplication against the CMR at the same time as the date feature. De-duplication is a more complicated feature and will likely need to be implemented in the HyP3 API rather than in DockerizedTopsApp. The date feature is more valuable and we should focus our efforts on one thing at a time.

PR #207, opened 2025-05-27
Update isce2_topsapp.__main__.gunw_slc to accept either `--reference-scenes` or `--reference-date`

asjohnston-asf:
Here's an alternative to https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/pull/205 for allowing on-demand users to specify a date instead of a list of scenes when ordering GUNW products. This eliminates the possibility that the user might not specify all SLCs necessary to fully cover the frame and thus produce a non-standard GUNW product.

This approach should have no impact on the ARIA team continuing to generate GUNW products using their own enumeration strategy and specifying a specific list of reference/secondary scenes when running DockerizedTopsApp locally or through a JPL HyP3 deployment.

Users running DockerizedTopsApp on-demand through ASF HyP3 would be required to specify reference/secondary dates instead of reference/secondary scenes; see https://github.com/ASFHyP3/hyp3/pull/2772 and https://github.com/ASFHyP3/hyp3-sdk/pull/350 .

~I've left some TODOs that ASF can work on over the next day or two, but let me know if there are any concerns with the general approach.~


mgovorcin:
Hi @asjohnston-asf, thanks for the update. I do have a few concerns about potential issues that might arise from this PR.

Currently, the code does not enforce the spatial coverage requirement over the specified frame. For example, we skip GUNW generation if the downloaded SLCs do not fully cover the frame (it is common to find SLC missing once in a while). As it stands, ISCE2 will crop to the common area, which might result in a non-standard ARIA GUNW product.

Additionally, updating the hyp3_sdk and modifying the aria_gunw job submission process would break our existing workflows for large-scale job enumeration and submission. 

Lastly, I understood from our meetings that de-duplication is a key part of this new addition.

asjohnston-asf:
> Currently, the code does not enforce the spatial coverage requirement over the specified frame.

Can you provide some examples? I'd expect a GUNW that includes as much spatial coverage as there is S1 data available would qualify as standard. For example, should it be possible to produce a GUNW for a frame over the coast where there is an S1 acquisition over the land portion but no subsequent acquisition over the ocean?

> Additionally, updating the hyp3_sdk and modifying the aria_gunw job submission process would break our existing workflows for large-scale job enumeration and submission.

This implementation takes care _not_ to break the ARIA team's existing interfaces. The [INSAR_ISCE](https://github.com/ASFHyP3/hyp3/blob/develop/job_spec/INSAR_ISCE.yml) job type remains unchanged; both it and DockerizedTopsApp will happily continue to accept lists of reference and secondary scenes.

> de-duplication is a key part of this new addition

De-duplication is a desired feature, but it is independent and less-valuable than date-based scene selection. ASF is taking an agile approach and choosing to implement and deliver date-based scene selection first. Users can then use it and get value from it while we then proceed to implement de-deduplication.

isce2_topsapp/localize_slc.py
```diff
@@ -232,3 +223,27 @@ def download_one(resp):
...
+    results = asf.search(
+        dataset=asf.constants.DATASET.SENTINEL1,
+        processingLevel=asf.constants.PRODUCT_TYPE.SLC,
+        beamMode=asf.constants.BEAMMODE.IW,
+        polarization=[asf.constants.POLARIZATION.VV, asf.constants.POLARIZATION.VV_VH],
+        flightDirection=frame.orbit_direction,
+        relativeOrbit=list({frame.relative_orbit_number_min, frame.relative_orbit_number_max}),
+        intersectsWith=frame.geometry.wkt,
+        start=date_as_datetime - datetime.timedelta(minutes=5),
+        end=date_as_datetime + datetime.timedelta(days=1, minutes=5),
```
jtherrmann, line 245: There may be an edge case here in which only scenes from just before or after the provided date are found, but none on the provided date itself.

asjohnston-asf:
> Currently, the code does not enforce the spatial coverage requirement over the specified frame.

And would it be sufficient to check for 100% overlap of the SLC scenes at https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/dev/isce2_topsapp/localize_slc.py#L86, rather than any overlap? This would prevent non-standard incomplete ifgs from being created, regardless of whether the inputs were specified via dates or lists of scenes.

asjohnston-asf:
Best documentation I can find is at https://github.com/ACCESS-Cloud-Based-InSAR/s1_frame_enumerator?tab=readme-ov-file#subtlties-of-creating-an-image-stack :
> 2. A specific frame has low coverage at numerous dates being at the boundary of a product distribution boundary (e.g. US and Mexico where Sentinel-1 might switch modes) or near a coastline - remove frame at the boundary or lower per frame or per pass coverage ratios

isce2_topsapp/localize_slc.py
asjohnston-asf: Agreed. I think it'll be sufficient to check that at least one of the search results matches the given date.

mgovorcin:
> Can you provide some examples? I'd expect a GUNW that includes as much spatial coverage as there is S1 data available would qualify as standard. For example, should it be possible to produce a GUNW for a frame over the coast where there is an S1 acquisition over the land portion but no subsequent acquisition over the ocean?

Here's [Charlie’s notebook](https://github.com/ACCESS-Cloud-Based-InSAR/s1_frame_enumerator/blob/dev/notebooks/Basic_Demo.ipynb) on job demo enumeration with fixed frames, and enumeration [workflow](https://github.com/ACCESS-Cloud-Based-InSAR/Aria-Hyp3-GUNW-Frame-Submission/blob/dev/1_Enumerating_A_Large_AOI.ipynb)

In short: we usually download 2 SLCs (reference and secondary) that fully cover the frame. ISCE2 clips them to the standard size using region_of_interest to ensure we get a standard GUNW product at the end.

Sometimes, one of the SLCs might be missing from ASF Vertex, like in early years 2015/2016. In that case, ISCE2 may receive 3 SLCs and will pick the area only with the overlapping bursts between reference and secondary, which in this case will only cover "half: the frame. We do use threshold, which you can find [here](url): 

> This implementation takes care _not_ to break the ARIA team's existing interfaces. The [INSAR_ISCE](https://github.com/ASFHyP3/hyp3/blob/develop/job_spec/INSAR_ISCE.yml) job type remains unchanged; both it and DockerizedTopsApp will happily continue to accept lists of reference and secondary scenes.

Great, thanks for clarifying that. I did notice the changes in  PR [src/hyp3_sdk/hyp3.py](https://github.com/ASFHyP3/hyp3-sdk/pull/350/files) to the `submit_aria_s1_gunw_job`, so I assumed this would break the current workflow.

> De-duplication is a desired feature, but it is independent and less-valuable than date-based scene selection. ASF is taking an agile approach and choosing to implement and deliver date-based scene selection first. Users can then use it and get value from it while we then proceed to implement de-deduplication.

I do know that David wants this as part of the feature, so might be best to discuss it on our next meeting.

asjohnston-asf:
Now that I read more closely, I see the notebook text suggests checking for 25% coverage, but the code block only checks for 10% coverage. What threshold do we want to enforce? Checking for 10% coverage isn't meaningfully different from the existing check for >0% coverage, in my opinion.,

isce2_topsapp/localize_slc.py
jtherrmann: Is it still worth attempting to handle this edge case? It seems like with a high enough `--min-frame-coverage` value then it shouldn't be an issue, but maybe we don't want to rely on that?

isce2_topsapp/localize_slc.py
```diff
@@ -238,3 +229,33 @@ def download_one(resp):
...
     }
+
+
+def _get_frame_by_id(frame_id: int) -> gpd.GeoSeries:
+    frames = gpd.read_file(Path(__file__).parent / 'data' / 's1_frames_latitude_aligned.geojson.zip')
+    return frames[frames.frame_id == frame_id].reset_index(drop=True).iloc[0]
+
+
+def _get_dates(product: asf.ASFProduct) -> set[datetime.date]:
+    date_strings = [product.properties[field].strip('Z') for field in ['startTime', 'stopTime']]
```
jtherrmann, line 240: If I delete `startTime` from this list, the test fails at the assertion on L192, as expected. But if I delete `stopTime`, then the test still passes. So it might be good to find a test case for crossing midnight where only `stopTime` falls on the given date.

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.6.0]
+
+### Added
+* `isce2_topsapp.__main__.gunw_slc` now accepts `--reference-date` and `--secondary-date` parameters as alternatives to
+  `--reference-scenes` and `--secondary-scenes`. When provided, `gunw_slc` will process all Sentinel-1 IW SLC scenes for
+  the given date and `--frame-id` in order to generate a standard ARIA_S1_GUNW product.
```
jtherrmann, line 14: Maybe we should avoid implying that using the date inputs is sufficient to achieve a standard product? Since there are other criteria (e.g. `--min-frame-coverage`) that must also be met.
```suggestion
  the given date and `--frame-id` in order to generate an ARIA_S1_GUNW product.
```

jtherrmann (dismissed):

jtherrmann (dismissed):

jtherrmann (approved):

PR #208, opened 2025-05-29
implement minimum frame coverage requirement

asjohnston-asf:
@mgovorcin this implements the 95% coverage requirement from your draft PR. 95% coverage seems like a steep requirement, though. As noted in https://github.com/ACCESS-Cloud-Based-InSAR/s1_frame_enumerator?tab=readme-ov-file#subtlties-of-creating-an-image-stack this would make it impossible to generate GUNW products over some areas such as the alleutian islands.

Does the ARIA team vary your coverage requirement depending on your area of interest? Do we need to leave the ARIA team the ability to dictate the minimum overlap required for their jobs, while restricting public on-demand users to a specific minimum overlap?

### DIFF
--- a/isce2_topsapp/localize_slc.py
+++ b/isce2_topsapp/localize_slc.py
@@ -57,7 +57,7 @@ def get_session():
 
 
 def get_interferogram_geo(
-    reference_obs: list, secondary_obs: list, frame_id: int = -1
+    reference_obs: list, secondary_obs: list, frame_id: int = -1, min_frame_coverage: float = 0.95,
 ) -> GeometryCollection:
     reference_geos = [shape(r.geojson()["geometry"]) for r in reference_obs]
     secondary_geos = [shape(r.geojson()["geometry"]) for r in secondary_obs]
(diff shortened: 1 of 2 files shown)


isce2_topsapp/localize_slc.py
```diff
@@ -83,9 +80,10 @@ def get_interferogram_geo(
     # Update the area of interest based on frame_id
     if frame_id != -1:
         gunw_geo = get_gunw_extent_from_frame_id(frame_id)
-        if not gunw_geo.intersects(ifg_geo):
+        frame_coverage = gunw_geo.intersection(ifg_geo).area / gunw_geo.area
+        if frame_coverage < min_frame_coverage:
             raise ValueError(
```
mgovorcin, line 85: It would be helpful to raise an error indicating whether the reference or secondary date fails to meet the spatial coverage requirement, to prevent submitting a date with missing coverage or gaps (e.g., due to a MultiPolygon).

jtherrmann (dismissed):

asjohnston-asf:
Agreed, that would be valuable. I'll merge this as-is, but would welcome a separate PR to update the messaging. We'll have that as a candidate on the to-do list at tomorrow's meeting if neither of ASF/ARIA have implemented it by then.

mgovorcin:
> Agreed, that would be valuable. I'll merge this as-is, but would welcome a separate PR to update the messaging. We'll have that as a candidate on the to-do list at tomorrow's meeting if neither of ASF/ARIA have implemented it by then.

Sounds good to me—thanks! I’ll go ahead and merge the PR to close it.

PR #221, opened 2025-09-17
Update ISCE2 to 2.6.4, python 3.11 + 3.12 support, able to use S1C data

cmarshak:


cmarshak:
@cmspeed @gracebato @jhkennedy - At some point today, I am going to upload a standard product that is generated by this PR that updates to the latest ISCE2.

I would encourage you to make sure the coseismic is nominally working too. We can likely merge this and you can just submit a single job to coseismic - or you could run locally - you just have to update the environment appropriately.

PR #228, opened 2025-10-13
upgrade to asf_search>=10.0.5

asjohnston-asf:
@cmarshak @sssangha asf_search issue https://github.com/asfadmin/Discovery-asf_search/issues/396 introduced sometime between Sep 26 and Oct 1 is causing all GUNW jobs in production to fail. This patch should be sufficient to resolve the issue. I'll verify once we have it merged to `dev` and I'll open a release PR promptly after that.


cmarshak (dismissed):

asjohnston-asf:
Also updated two GUNW download URLs to resolve two failing notebook tests.

cmarshak (approved):

PR #240, opened 2026-03-06
Add CDSE endpoint for data download

dbekaert:
Kept all logic in place for checking meta-data at ASF, however added an option when downloading the actual granule to fetch it from CDSE. This might make the plug-in more accessible for European users or those that are operating in the European cloud ecosystem. 

Pointing the the compressed CDSE S1 archive above the uncompressed to limit download traffic. ISCE can handle the virtual access to the compressed files.

### DIFF
--- a/isce2_topsapp/__main__.py
+++ b/isce2_topsapp/__main__.py
@@ -257,7 +257,7 @@ def gunw_slc():
     args = parser.parse_args()
     args = update_slc_namespace(args)
 
-    # Validation 
+    # Validation
     ensure_earthdata_credentials(args.username, args.password)
     # Also validate CDSE credentials if using CDSE for download
     if args.download_source == "CDSE":
--- a/isce2_topsapp/__main__.py
+++ b/isce2_topsapp/__main__.py
@@ -257,7 +257,7 @@ def gunw_slc():
     args = parser.parse_args()
     args = update_slc_namespace(args)
 
-    # Validation - always need Earthdata creds for ASF metadata lookup
+    # Validation 
     ensure_earthdata_credentials(args.username, args.password)
     # Also validate CDSE credentials if using CDSE for download
     if args.download_source == "cdse":
--- a/isce2_topsapp/__main__.py
+++ b/isce2_topsapp/__main__.py
@@ -44,6 +44,7 @@ def localize_data(
     dry_run: bool = False,
     water_mask_flag: bool = True,
     geocode_resolution: int = 90,
+    download_source: str = "asf",
 ) -> dict:
     """The dry-run prevents gets necessary metadata from SLCs and orbits.
 
@@ -58,6 +59,7 @@ def localize_data(
         frame_id=frame_id,
         min_frame_coverage=min_frame_coverage,
         dry_run=dry_run,
+        download_source=download_source,
     )
 
     out_orbits = download_orbits(reference_scenes, secondary_scenes)
@@ -204,6 +206,18 @@ def get_slc_parser():
         default=0.5,
         help="The power applied to the patch FFT of the phase filter",
     )
+    parser.add_argument(
+        "--download-source",
+        type=str,
+        choices=["asf", "cdse"],
+        default="asf",
+        help=(
+            "Source for downloading Sentinel-1 SLC granules. "
+            "'asf' uses Alaska Satellite Facility (requires Earthdata credentials). "
+            "'cdse' uses Copernicus Data Space Ecosystem (requires CDSE credentials "
+            "via CDSE_USERNAME/CDSE_PASSWORD env vars or ~/.netrc)."
+        ),
+    )
     return parser
 
 
@@ -243,12 +257,16 @@ def gunw_slc():
     args = parser.parse_args()
     args = update_slc_namespace(args)
 
-    # Validation
+    # Validation - always need Earthdata creds for ASF metadata lookup
     ensure_earthdata_credentials(args.username, args.password)
+    # Also validate CDSE credentials if using CDSE for download
+    if args.download_source == "cdse":
+        from isce2_topsapp.localize_slc_cdse import ensure_cdse_credentials
+        ensure_cdse_credentials()
     cli_params = vars(args).copy()
     [
         cli_params.pop(key)
-        for key in ["username", "password", "bucket", "bucket_prefix", "dry_run"]
+        for key in ["username", "password", "bucket", "bucket_prefix", "dry_run", "download_source"]
     ]
     topsapp_params_obj = topsappParams(**cli_params)
 
@@ -268,6 +286,7 @@ def gunw_slc():
         frame_id=args.frame_id,
         min_frame_coverage=args.min_frame_coverage,
         water_mask_flag=args.estimate_ionosphere_delay,
+        download_source=args.download_source,
     )
     loc_data["frame_id"] = args.frame_id
     loc_data["cmd_line_str"] = cmd_line_str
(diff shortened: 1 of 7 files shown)


dbekaert:
@copilot update the changelog

dbekaert:
@copilot update the README to include an example for the CDSE download option. Please review the existing readme and follow the structure already established

dbekaert:
@copilot can you check the failed [call-ruff-workflow / check-with-ruff](https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/actions/runs/22782376263/job/66091141392?pr=240#logs)
failed 2 minutes ago in 3m 4s. and make the necessary corrections?

isce2_topsapp/__main__.py
```diff
@@ -204,6 +206,18 @@ def get_slc_parser():
         default=0.5,
         help="The power applied to the patch FFT of the phase filter",
     )
+    parser.add_argument(
+        "--download-source",
+        type=str,
+        choices=["asf", "cdse"],
```
jhkennedy: This is admittedly a total nit-pick and not blocking, but I prefer uppercase here, like:
```
choices=["ASF", "CDSE"]

File: isce2_topsapp/__main__.py [hunk 2]
```diff
@@ -243,12 +257,24 @@ def gunw_slc():
     args = parser.parse_args()
     args = update_slc_namespace(args)
 
-    # Validation
+    # Validation - always need Earthdata creds for ASF metadata lookup
```
jhkennedy: 🤔 do we? I don't think we point to any restricted collections so unauthenticated CMR searches should work.

jhkennedy (commented):
I did a quick, cursory review, and it looks good to me so far. @cmarshak do you have time for a proper review? @dbekaert I'm swamped for the next 3 weeks, so I'm not sure I'll get to it before that. But it's probably just fine to send to `dev`.

isce2_topsapp/__main__.py
dbekaert: true i think for meta-data that we dont need authentication, rather only when we start downloading.

dbekaert:
> I did a quick, cursory review, and it looks good to me so far. @cmarshak do you have time for a proper review? @dbekaert I'm swamped for the next 3 weeks, so I'm not sure I'll get to it before that. But it's probably just fine to send to `dev`. 

Thank @jhkennedy. @cmarshak any chance we could get this in over by early next week?

isce2_topsapp/__main__.py
cmarshak: What about the option of "both"?

isce2_topsapp/localize_slc_cdse.py
```diff
@@ -0,0 +1,350 @@
...
+            if out_path.stat().st_size > 0:
+                print(
+                    f"Successfully downloaded {out_filename} from CDSE ({out_path.stat().st_size / 1e6:.1f} MB)"
+                )
+                return out_filename
+
+            print("Downloaded file is empty, retrying...")
+            out_path.unlink(missing_ok=True)
+
+        except requests.RequestException as e:
```
cmarshak: Use tenacity please - much nicer.

Requests has something, but this will be the most robuist and easy to read.

An example is here: https://github.com/asfadmin/Discovery-asf_search/blob/03ec63a31d02c44e53140c8b0f172e6de3e158fd/asf_search/search/search_generator.py#L347

isce2_topsapp/__main__.py
cmarshak: For metadata, no authentication. For download, we do.

isce2_topsapp/localize_slc_cdse.py
cmarshak: https://tenacity.readthedocs.io/en/latest/

cmarshak (dismissed):

cmarshak:
@dbekaert 🔥🔥🔥

Nice work! Thanks for doing this!

I put some general comments in PR. We can circle back:

- should be asf, csde, or both, right? Like try one and then the other?
- also suggested you use tenacity for retries - super useful and very nice to read - removes retry logic from function
- please remove unnecessary print statemetnts. I know LLMs like that, but seems unnecessary to me.

Thank you for doing this! Looks great.

I am on vacation Mon-Weds next week so if Joe doesn't want approve we can merge later.

Also, why do you want to merge? We could quickly merge to dev even while I am gone because `dev` will not be production. Won't have a chance to check the example you share until I am back.

One think I really liked - using the auth token outside the multithreaded download. Think that effectively creates a single session for all the downloads!

isce2_topsapp/localize_slc_cdse.py
dbekaert: @copilot please see if you can use tenacity here.

isce2_topsapp/__main__.py
dbekaert:
```suggestion
    # Validation 
```

isce2_topsapp/__main__.py
dbekaert: I think that would be nice, but not critical at the moment. would add this to the todo list

isce2_topsapp/__main__.py
dbekaert: this has been updated.

isce2_topsapp/__main__.py
dbekaert: @copilot  please change the "choices" of to be using capitalized ASF and CDSE instead of asf and cdse. Please flow throughhe code to ensure this is passed correctly from the parser options down in the code.

dbekaert:
> @copilot can you check the failed [call-ruff-workflow / check-with-ruff](https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/actions/runs/22782376263/job/66091141392?pr=240#logs) failed 2 minutes ago in 3m 4s. and make the necessary corrections?

@copilot  can you asses again: https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/actions/runs/23191828624/job/67389676299?pr=240

dbekaert:
@cmarshak @jhkennedy  this should be addressing the issues that were raised. I will do one more test to create end-2-end with ASF and CDE option.

dbekaert:
upon testing i realized that the natively compressed archive is not always available as its only <1m rolling. hence the approach is to fallback to the compressed archive which has the complete record be it at a cost the zip files are larger.

dbekaert:
@copilot check the [Static Analysis (Flake8/Ruff) / call-ruff-workflow / check-with-ruff (push)](https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/actions/runs/23207624709/job/67447356850?pr=240)Failing after 3m

dbekaert:
@cmarshak @jhkennedy I confirm all is still working and download for both the CDSE and ASF portions started. Those where the changes were made in the code.

I also confirm the fallback for CDSE is working between the different archives:
```
/home/jovyan/.local/envs/topsapp_env/lib/python3.11/site-packages/scipy/__init__.py:155: UserWarning: A NumPy version >=1.18.5 and <1.26.0 is required for this version of SciPy (detected version 1.26.4
  warnings.warn(f"A NumPy version >={np_minversion} and <{np_maxversion}"
Downloading SLCs from CDSE:   0%|                         | 0/4 [00:00<?, ?it/s]CDSE download attempt #1 for S1A_IW_SLC__1SDV_20220212T222803_20220212T222830_041886_04FCA3_2B3E
CDSE download attempt #1 for S1A_IW_SLC__1SDV_20220131T222803_20220131T222830_041711_04F690_8F5F
CDSE download attempt #1 for S1A_IW_SLC__1SDV_20220131T222828_20220131T222855_041711_04F690_28D7
CDSE download attempt #1 for S1A_IW_SLC__1SDV_20220212T222828_20220212T222855_041886_04FCA3_A3E2
Compressed format not available, falling back to uncompressed...
Compressed format not available, falling back to uncompressed...
Compressed format not available, falling back to uncompressed...
Compressed format not available, falling back to uncompressed...
```


started the download:

```
drwxr-xr-x. 7 jovyan jovyan 6.0K Mar 17 18:27 ..
drwxr-xr-x. 3 jovyan jovyan 6.0K Mar 17 18:28 .
-rw-r--r--. 1 jovyan jovyan 659M Mar 17 18:28 S1A_IW_SLC__1SDV_20220131T222803_20220131T222830_041711_04F690_8F5F.zip
-rw-r--r--. 1 jovyan jovyan 195M Mar 17 18:28 S1A_IW_SLC__1SDV_20220212T222828_20220212T222855_041886_04FCA3_A3E2.zip
-rw-r--r--. 1 jovyan jovyan 976M Mar 17 18:28 S1A_IW_SLC__1SDV_20220212T222803_20220212T222830_041886_04FCA3_2B3E.zip
-rw-r--r--. 1 jovyan jovyan  59M Mar 17 18:28 S1A_IW_SLC__1SDV_20220131T222828_20220131T222855_041711_04F690_28D7.zip
```

cmarshak (approved):

